Issue #70, opened 2014-09-10
Missing iwinfo


basros:
Using `iw` is a PITA, `iwinfo` would reduce that.

PR #71, opened 2014-09-11
[packages] Add iwinfo to list

cholin:
output of iw is verbose. For sitesurvey, channel choosing or
station dump iwinfo is more human friendly to read. see #70

### DIFF
--- a/packages/minimal.txt
+++ b/packages/minimal.txt
@@ -21,6 +21,7 @@ kmod-ipip
 kmod-iptunnel4
 qos-scripts
 firewall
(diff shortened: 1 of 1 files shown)


SvenRoederer:
@booo iwinfo seems still to be used by the assistent: https://github.com/freifunk-berlin/firmware-packages/blob/master/utils/luci-app-ffwizard-berlin/luasrc/model/cbi/freifunk/assistent/wireless.lua#L312
not sure if this also applies to the commandline-tool you like to remove from the image

PR #195, opened 2015-01-06
Fix redirect after first boot

booo:
I suggest we fix the first part of the redirect problem this way. @geirkairam what do you think?


booo:
Works for me on WDR4900. @geirkairam can you comment on my changes. Are you fine with them? If so please merge.

Issue #191, opened 2014-12-24
no space left - after initial flashing of wrt841nd v9 (Firmware Berlin (kathleen 0.0.0, based OpenWRT r42554)


research7:
Any hints on how to resolve this (description below)?
Thanks!

Firmware: Firmware Berlin (kathleen 0.0.0, based OpenWRT r42554) (downloaded yesterday)
I cannot update certain settings in the web interface, i.e. these updates wont be saved even though
a successful commit was claimed.
For example I attempted to add my geolocation to the basic settings, but I cannot save these.
Neither can I add a wifi access point or change settings for the existing interfaces.
At least I could enable ssh access.
Write attempts to anything other than /tmp were not successful due to "no space left" errors.

This is a "df -h" output:
Filesystem                Size      Used Available Use% Mounted on
rootfs                  320.0K    264.0K     56.0K  83% /
/dev/root                 2.8M      2.8M         0 100% /rom
tmpfs                    14.3M    728.0K     13.5M   5% /tmp
/dev/mtdblock3          320.0K    264.0K     56.0K  83% /overlay
overlayfs:/overlay      320.0K    264.0K     56.0K  83% /
tmpfs                   512.0K         0    512.0K   0% /dev

booo:
We can't do much about this and **I** don't want to do much about it. **I** suggest you buy proper hardware with at least 8MB of flash memory.

You can try to find some packages that are superfluous and we can remove them if it makes sense.

research7:
Thanks for the reply.
Well, according to the "df -h" output I posted above there is 13.5MB free memory. 
Hmm? Where's my mistake?

I do have a  wrt841nd v9 model and downloaded the recommended firmware from
freifunk-berlin.

booo:
13.5MB on the temporary file sysstem. Would you save configuration on a temporary file system?

I don't the you did something wrong. It's just too small...

research7:
I agree, that saving important data on temp fs is not a great idea. but, I assumed that 
this could maybe be solved by sort of repartitioning or whatever ... 

I'll look at your suggestion about finding superfluous packages.

Is there a way to file a bug to the firmware-berlin distribution site?
I think it might be useful to tell people before downloading and flashing 
to check the memory available on the router. That would also avoid bug reports here.

Thanks!

booo:
I don't think you can repartition. You can find more on the flash layout etc. [here](http://wiki.openwrt.org/doc/techref/flash.layout).

[...]

Issue #52, opened 2014-08-20
remove freifunk-watchdog


booo:
The functionality of the freifunk-watchdog package overlaps at some points with functionality provided by procd. Maybe we should remove the package or update the source code.

http://luci.subsignal.org/trac/browser/luci/trunk/contrib/package/freifunk-watchdog

booo:
There is no procd support yet :/ We can have a look at this. Shouldn't be to complex. Working draft is here:

https://github.com/booo/openwrt-routing/tree/olsrd_procd

Should we wait with the removal till we implemented support for procd in the olsr package?

lgierth:
Updated olsrd+procd tree here (WIP): https://github.com/lgierth/openwrt-routing-packages/compare/olsrd_procd

cholin:
To get rid of `freifunk-watchdog` we have to break up this whole legacy dependency chain in luci. `freifunk-watchdog` gets selected by `luci-mod-freifunk-community`. This package itself is a meta-package (no content) and selects the following packages:

```
+iptables-mod-nat-extra +iptables-mod-ipopt +luci-app-splash \
+olsrd +olsrd-mod-dyn-gw-plain +olsrd-mod-jsoninfo +olsrd-mod-nameservice \
+olsrd-mod-watchdog +kmod-tun +ip +freifunk-watchdog +luci-app-olsr \
+luci-app-olsr-services +freifunk-gwcheck +freifunk-mapupdate
```

We do not use packages like:

* `freifunk-gwcheck`
* `freifunk-mapupdate`
* `luci-app-splash`
* `olsrd-mod-watchdog`

In general this meta package does not make much sense to us anymore because we have our packages lists in `packages/` for that. I would favor to get rid of it instead of patching the meta-package and explicitly mention these packages in `packages/*.txt`. This would mean as well that we don't use `luci-mod-freifunk` as well. The benefits of this would to reduce complexity of dependency chains, get rid of legacy code and save space (a side note: Gluon did the same).

andrenarchy:
+1 for explicit package list!

lynxis:
+1

sarumpaet:
[...]

PR #203, opened 2015-01-18
strip firmware size

cholin:
As there are routers with 4MB in the wild and we do offer firmwares on http://berlin.freifunk.net/downloads/ I would be happy if we could reduce the size for these devices. With this pull request I added the following changes:

* remove `iwinfo` and `luci-app-owm-gui` packages for 4MB devices
* remove `obsolete freifunk-gwcheck` package (this functionality has been replaced with olsrd gateway check plugin)
* remove ssp support
* activate `CONFIG_STRIP_KERNEL_EXPORTS` (unnecessary / unlinked functions in the kernel itself will be removed)

Support for stack smashing protection got merged with 6318d5a. As there is no package which supports SSP at the moment in the Freifunk Firmware Berlin this functionality is obsolete and can be deleted for size and complexity reasons (see #107 and #191 (comment)).

I tested this with a TP-Link 841n v8 and got 48KB. Before there was 96KB of free space and afterwards 144KB.


lgierth:
I agree that SSP should be removed.

cholin:
With the last changes I got  156KB (additional 12KB) on a TP-LINK 741:

```
root@OpenWrt:~# df -h
Filesystem                Size      Used Available Use% Mounted on
rootfs                  384.0K    228.0K    156.0K  59% /
/dev/root                 3.0M      3.0M         0 100% /rom
tmpfs                    14.4M    172.0K     14.2M   1% /tmp
tmpfs                    14.4M     88.0K     14.3M   1% /tmp/root
tmpfs                   512.0K         0    512.0K   0% /dev
/dev/mtdblock3          384.0K    228.0K    156.0K  59% /overlay
overlayfs:/overlay      384.0K    228.0K    156.0K  59% /
```

Hard to test which change are worth the effort because I don't have a device on my own here. How do we want to proceed? Is anyone interested into getting (some of) this changes upstream to better support these 4MB devices?

booo:
I have a spare 841ND if you need some test hardware.

To me the consequences of stripping the kernel and changing the block size of the squashfs are not clear.

If you really thing this is useful let's merge it!

cholin:
I think we should evaluate each commit and than think again if we want this for our target `ar71xx`. Some commits may do not result in that big amount of free space or limit us in another way. Some words on kernel stripping and block size for squashfs:

* kernel stripping: not needed kernel functionality will be removed. This may result in non-functional behaviour for kernel modules which got installed later via opkg (see as well https://forum.openwrt.org/viewtopic.php?pid=182963#p182963)

* squash fs blocksize: Higher block sizes for squashfs will result in smaller space consumption. The reason for that is the bigger the block size the better the compression and therefor smaller partition size.

cholin:
Seems I forgotten to paste `df -h` output of https://github.com/freifunk-berlin/firmware/commit/b9f3531aa548ff7073c5380dfeeca869ec070c7d:

```
root@OpenWrt:~# df -h
Filesystem                Size      Used Available Use% Mounted on
rootfs                  448.0K    228.0K    220.0K  51% /
/dev/root                 2.5M      2.5M         0 100% /rom
tmpfs                    14.4M    160.0K     14.2M   1% /tmp
tmpfs                    14.4M     88.0K     14.3M   1% /tmp/root
tmpfs                   512.0K         0    512.0K   0% /dev
/dev/mtdblock3          448.0K    228.0K    220.0K  51% /overlay
overlayfs:/overlay      448.0K    228.0K    220.0K  51% /
```

It results in additional 64KB (in total 220KB).

research7:
Maybe, my experience using the patch on a wrt841nd v9 is of some help. thus, I repost some info from #191:

Works on wrt841nd
===
I applied the suggested changes to master, build the image, flashed, and things work fine now!
Addition
===
in order to make the Freifunkassitent work I had to scp profile_berlin to /etc/config


Space left after successful flashing and configuration
===
Filesystem Size Used Available Use% Mounted on
rootfs 384.0K 236.0K 148.0K 61% /
/dev/root 2.5M 2.5M 0 100% /rom
tmpfs 14.3M 176.0K 14.2M 1% /tmp
/dev/mtdblock3 384.0K 236.0K 148.0K 61% /overlay
overlayfs:/overlay 384.0K 236.0K 148.0K 61% /
tmpfs 512.0K 0 512.0K 0% /dev

sarumpaet:
On a WR841N v8, I got this:
Filesystem Size Used Available Use% Mounted on
overlayfs:/overlay      448.0K    220.0K    228.0K  49% /
Doing some "dd if=/dev/urandom" in the overlay was okay up to 3x64k, then the space was exhausted, but deleting files in the overlay still worked (in contrast to kathleen 0.0.0).

So, well done!

I also tried without CONFIG_STRIP_KERNEL_EXPORTS, but this reduces the overlay size from 440k to 384k and free space below the 3x64k mark.

Would it be possible to have a special kernel configuration for 4MB devices only?

cholin:
This PR is not up-to-date anymore and contains changes we may do not want. I close this for now and recommend to focus on https://github.com/freifunk-berlin/firmware/issues/52#issuecomment-75763624 instead.

cholin:
Please do not delete this branch. I may cherry-pick some commits for another PR.

lgierth:
Have you tried the following? It helped me get the cjdns image below the 3473408 bytes mandated by the Netgear's factory partitioning.

```
CONFIG_STRIP_KERNEL_EXPORTS=y # Strip unnecessary exports from the kernel image
CONFIG_USE_MKLIBS=y # Strip unnecessary functions from libraries
```

lgierth:
Nevermind, you did.

cholin:
Yeah. Already included (see commit list).

PR #204, opened 2015-01-19
[packages] add luci-app-firewall

cholin:
At the moment users can't change firewall settings without using a shell. This
adds the `luci-app-firewall` package to empower people without shell knowledge to
change firewall settings on their own.


booo:
Looks good. Thanks! I wonder why we did not add this in the first place.

Issue #39, opened 2014-08-14
Support for different packages lists


cholin:
At the moment there is only `packages/minimal.txt` as package list for a firmware. I think the Makefile should support arbitrary amount of package lists files like the old script did. Then we could generate different firmwares like for VPN or BBB.

cholin:
Added functionality for this with #207.

Issue #144, opened 2014-10-27
hotplug script for nodes without ffvpn


booo:
If you want to provide internet without the ffvpn you have to copy the default route from the main table to the olsr-tunnel table. Attatched is a unfinished script to do so via netifds hotplug capability:

``` sh
#!/bin/sh

. /lib/functions.sh
. /lib/functions/network.sh

logger -s -t fooo $@

local SCRIPT_NAME='dummy' #TODO fix this
local TABLE='olsr-tunnel'
local default_gateway


logger -s -t $SCRIPT_NAME "INTERFACE: ${INTERFACE} (DEVICE: ${DEVICE}) ACTION: ${ACTION}"
#TODO check what ACTIONS exists
if [ "${ACTION}" = "ifup" ]; then
    if [ "${INTERFACE}" = "freifunk" ]; then
        # default via 104.205.0.73 dev br-freifunk  proto static 
        default_gateway=$(ip route show table main | grep default | cut -d ' ' -f 3)

        logger -s -t $SCRIPT_NAME ">${default_gateway}<"

        if [ ! -z "${default_route}" ]; then
            return
        fi
        #TODO check for error with $? ???
        #insert default route in olsr-tunnel
        ip route add default via $default_gateway dev ${DEVICE} table ${TABLE}
        #remove default route from main
        #ip route del default via $default_gateway dev ${DEVICE} table main
    fi
fi

if [ "${ACTION}" = "ifdown" ]; then
    if [ "${INTERFACE}" = "freifunk" ]; then
        default_gateway = $(ip route show table ${TABLE} | grep default | cut -d ' ' -f 3)
    if [ ! -z "${default_gateway}" ]; then
            return
    fi
        ip route del default via $default_gateway dev ${DEVICE} table ${TABLE}
    fi
fi

```

booo:
If you do so you should change the `ping` command in /etc/config/olsrd as well.

cholin:
I think this should be implemented in `/etc/hotplug.d/iface/30-policyrouting`. If policyrouting isn't enabled or strict is set to false, then the default route should as well added to `olsr-tunnel` as well. The PingCmd of olsrd should be outsourced into a script. We could use the same conditions there to either ping through ffvpn or not.

SvenRoederer:
superseded by https://github.com/freifunk-berlin/firmware-packages/pull/125 (referenced in https://github.com/freifunk-berlin/firmware/issues/212#issuecomment-340233463)

PR #207, opened 2015-01-23
Features/multiple packages lists

cholin:
This PR is open for comments. It includes the following changes:

* Initial support for multiple packages lists (see #39)
* add basic packages list for backbone nodes (if there is no 4MB suffix pacakge list file it will use the original list for all profiles)

The resulting directory structure for firmwares looks like the following:

```
firmwares/
  bbb/
     TARGET/
       ImageBuilder
       Images
  vpn/
     TARGET/
       ImageBuilder
       Images
  packages/
    TARGET/
       base/
       luci/
       packages/
       packages_berlin/
       routing/
```

These changes do as well affect our buildbot infrastructure.


packages/bbb.txt
```diff
@@ -0,0 +1,68 @@
...
+
+# Common
+mtr
+iperf
+tmux
+vnstat
+ethtool
+dnsmasq
+iwinfo
+libiwinfo-lua
```
booo: I suggest we add tcpdump here.

cholin: done

booo:
We should add some documentation to the README.

I have to take a closer look but why do we have the image builder twice per target? We only have to create different images/package sets with the image builder per target, right?

cholin:
Well we could as well create a directory for these image builders like for `packages`. But I think for clearness I just adapted the old directory structure (with IB inside the dir with the images). It is as well the same layout which OpenWRT itself uses.

andrenarchy:
Just talked to @cholin and we think it's a good idea to change it to the following layout:
```
firmwares/
  TARGET/
    BUILDBOT_NUMBER/      // only for unstable builds
      OpenWrt-ImageBuilder-....tar.bz2
      bbb/
         Images
      vpn/
         Images
      packages/
         base/
         luci/
         packages/
         packages_berlin/
         routing/
```
Opinions?

cholin:
@andrenarchy @booo can you test this PR and give some feedback? :smile:

andrenarchy:
Testing right now. Regarding the discussion in #212 and #144, I suggest we rename `vpn` to `default` (or some other neutral term).

cholin:
Maybe we should rename both packages list in the following matter:

* `vpn` => `default`
* `bbb` => `minimal`

What do you think?

cholin:
As @booo proposed: Rebased on top of https://github.com/freifunk-berlin/firmware/pull/211.

andrenarchy:
btw: the build failed due to source files that could not be downloaded (see [ar71xx build 479](http://buildbot.berlin.freifunk.net/builders/ar71xx/builds/479)). Maybe the rebase fixed it. If [ar71xx build 481](http://buildbot.berlin.freifunk.net/builders/ar71xx/builds/481) succeeds, I'll also test it locally again.

booo:
I find the name "minimal" misleading. I wouldn't expect for example these packages in a minimal set:
https://github.com/freifunk-berlin/firmware/pull/207/files#diff-37b55f51476a6204d626944e1769b02dR18

"bbb" was fine to me. "expert" would be one suggestions but that's strange too.

andrenarchy:
@booo: yeah, `minimal` may be misleading. But the abbreviation `bbb` may be unclear. `expert` or `advanced` could be an alternative...

cholin:
[ar71xx build 481](http://buildbot.berlin.freifunk.net/builders/ar71xx/builds/481) finished successfully. See [buildbot/unstable/ar71xx/481](http://buildbot.berlin.freifunk.net/buildbot/unstable/ar71xx/481/) for the directory layout.

I agree that `minimal` may be the wrong label. Maybe just `backbone` or `manual`? With `expert` and `advanced` some may expect additional features but quite the contrary is true (there is no wizard at all...).

andrenarchy:
Compilation also works fine here and results in the intended directory structure.

I suggest we merge this PR. Note: it contains the changes from #211 which have been tested by @booo. So #211 can be closed after merging this one.

booo:
I suggest that we document the structure in the README

I vote for backbone as name for "minimal".

We still need to update the buildbot configuration.

cholin:
Added some docu and renamed `minimal` to `backbone`.

andrenarchy:
What exactly needs to be changed in the buildbot config? The latest commit builds and is uploaded to the download location:

 * [ar71xx](http://buildbot.berlin.freifunk.net/buildbot/unstable/ar71xx/486/)
 * [ar71xx_mikrotik](http://buildbot.berlin.freifunk.net/buildbot/unstable/ar71xx_mikrotik/429/)
 * [mpc85xx](http://buildbot.berlin.freifunk.net/buildbot/unstable/mpc85xx/437/)
 * [x86](http://buildbot.berlin.freifunk.net/buildbot/unstable/x86/325/)

cholin:
@booo can you say something about buildbot? I think we could merge this anyhow now :smile:

sarumpaet:
For the record, b1fd7dc1ad17cf581e567bb4ec433fd9abb1c415 fixes a typo in 066e80e3523b0cdc2d47bf843c8d5dd8ff0980c2 that prevented suffixed package lists from being found/built.

PR #211, opened 2015-02-02
config.mk: Update OpenWrt revision to 44162

booo:
The openwrt-package-openvpn-cipher-none-regression.patch is obsolete because
it's now included upstream:

http://git.openwrt.org/?p=14.07/openwrt.git;a=commit;h=c2a8bebec3bbb7dbe9d3e9fa50614dab230eac6e


booo:
"Tested" on WDR4900. Seems to work.

andrenarchy:
@booo since you tested it, we should merge this PR.

Issue #155, opened 2014-11-01
create procd config for olsrd


slomo:
Let olsrd be managed by procd, so it doen't need supervision by the freifunk-watchdog anymore (#52).

cholin:
Seems you removed `START` from the init scripts. This is still needed. As far as I know you should define as well a `service_triggers()` function. With it the daemon will be restarted automatically if the uci config file gets modified (if md5 hash of the file  differs).

I would split some changes into their own commits. I already did this in [olsrd_procd](https://github.com/cholin/openwrt-routing-packages/commits/olsrd_procd). I will test it in the next days and try to merge it upstream.

cholin:
I talked to @booo and we agreed that maybe automatically restarting olsrd if you do a change to `/etc/config/olsrd` is not a good solution.

andrenarchy:
+1! You could easily be locked out of your router.

cholin:
I've tested a working procd integration with [cholin-openwrt-routing-packages/olsrd_procd](https://github.com/cholin/openwrt-routing-packages/commits/olsrd_procd) in a mini testbed consisting of a WDR3500, 3600 and bbbvpn connection to the mesh itself. So far it seems to work great. See as well https://github.com/openwrt-routing/packages/pull/78.

booo:
This ticket is about olsrd!

pmelange:
oops, thanks for keeping me honest.

SvenRoederer:
Since early 2015 there was no progress on this neighter from us nor from upstream (https://github.com/openwrt-routing/packages/pull/78). Honestly it seems nobody will invest time in this at all, so lets close this.

PR #223, opened 2015-02-17
olsrd: make use of procd, nuke freifunk-watchdog

lgierth:
See #52 and #155, upstreamed as openwrt-routing/packages#78

### DIFF
--- a/patches/016-olsrd-procd.patch
+++ b/patches/016-olsrd-procd.patch
@@ -1,0 +1,1886 @@
+From 0c25b956132a834c4a75e223723ac1cc66d2c556 Mon Sep 17 00:00:00 2001
+From: Nico Geyso <nico.geyso@fu-berlin.de>
+Date: Wed, 4 Feb 2015 11:49:52 +0100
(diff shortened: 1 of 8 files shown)


patches/016-olsrd-procd.patch
```diff
@@ -0,0 +1,1886 @@
...
++
++	procd_set_param respawn
++
++	procd_close_instance
++
++	olsrd_setup_smartgw_rules
++}
++
++service_triggers() {
++	procd_add_reload_trigger $OLSRD
```
cholin: I think restarting olsrd if the uci config changes is not a good default behaviour. We may should restart if uci network changes but this should be covered with `procd_add_reload_interface_trigger`

patches/016-olsrd-procd.patch
```diff
@@ -0,0 +1,1886 @@
...
```
cholin: see comment above

booo:
I'm closing this one. We should reopen or create a new pull requests once the procd configuration for olsrd is merged upstream

PR #225, opened 2015-02-19
fix out of tree build + speed up build

lynxis:


andrenarchy:
Why carambola2? Is this just any device so it doesn't build all the images in the `compile` target because the actual images are created within the `firmwares` target?

lynxis:
I like the carambola2 ;). But it doesn't matter which one, but most targets are multiboard targets.
> Is this just any device so it doesn't build all the images in the compile target because the actual images are created within the firmwares target?
Correct.

andrenarchy:
Alright. :smiley: I pushed the branch to this repo so the buildbot is triggered: https://github.com/freifunk-berlin/firmware/tree/lynxis/fastbuild

lynxis:
It works, but the imagebuilder is now missing some packages.
E.g. ohci package is missing for some profiles, because the carambola2 doesnt depend on that.
Also the imagebuilder doesn't fail. See #228.

cholin:
What is the status of this PR? Seems we need the `default` target, don't we?

lynxis:
no. It's missing some package, but still saves some time.

lynxis:
rebase against freifunk/master

lynxis:
Buildbot build was successful. See http://buildbot.berlin.freifunk.net/builders/ar71xx/builds/547

andrenarchy:
Thx @lynxis, merging in a minute...

PR #233, opened 2015-02-27
[Makefile] fix fail of imagebuilder doesn't fail the build

### LINKED ISSUE #228: [firmware] we don't check returncode of the imagebuilder
If one profile (e.g. ubnt) fails, the imagebuilder doesn't fail in all cases.
We need to check the return code for every imagebuilder's build.

lynxis:
In some cases the imagebuilder can fail on some profiles but
doesn't report back the fail to make. Fix #228.


lgierth:
Why did it not fail in the beginning if `make [..] image [..]` returns not-zero? Just curious as I might have to deal with this as well in the future.

lynxis:
Because the imagebuilder target isn't written in Make, it's written in bash.

lgierth:
Oh okay, thanks -- I usually run my sh in bash with `-e`, so I was just a little surprised :)

lynxis:
@lgierth do you know how we can enable `-e` with our Makefile?

andrenarchy:
The PR is fine but there's still a bug in imagebuilder that has to be fixed, see #228.

PR #234, opened 2015-02-28
Bugfix/remove migration from backbone

cholin:
Backbone installations are configured manually. If we run our migration script
here it will break the config. Therefor remove the migration script from package
list.

### DIFF
--- a/packages/backbone.txt
+++ b/packages/backbone.txt
@@ -5,7 +5,6 @@
 # Defaults
 freifunk-berlin-dhcp-defaults
 freifunk-berlin-freifunk-defaults
-freifunk-berlin-migration
 freifunk-berlin-network-defaults
 freifunk-berlin-olsrd-defaults
 freifunk-berlin-statistics-defaults


andrenarchy:
Why do we need iptables?

cholin:
Testing purposes and some backbone installation may need custom configuration including custom rules.

lynxis:
- remove migration can be merge.
Isn't iptables already included as a dependency by fw3?

cholin:
Nope. `fw3` selects the following packages:

```
Selects: PACKAGE_kmod-ipt-nat [=y] && PACKAGE_libc [=y] && PACKAGE_libuci [=y] && PACKAGE_libip4tc [=y] &&  PACKAGE_libpthread [=m] && PACKAGE_libubus [=y] && PACKAGE_kmod-ipt-core [=y] && PACKAGE_libxtables [=y] &&  PACKAGE_libip6tc [=y] && PACKAGE_librt [=m] && PACKAGE_libubox [=y]
```

But as we use *qos-scripts*, *iptables* should be already included. Therefore I removed `[packages] add iptables to backbone`.

lynxis:
LGTM

andrenarchy:
@cholin: usually the backbone sites don't have the firewall (no NAT required!) or QOS packages installed. Is the firewall activated by the default config? This may break backbone sites when upgrading from an image without firewall I guess.

booo:
Seems like the firewall is still included for whatever reason. I can't find qos in this image. But I think this is a different topic.

Pull request looks good to me. If you agree I merge this at the end of the day.

andrenarchy:
I'm also happy with the PR. Since everyone is satisfied, I'll merge it.

The qos/firewall issue is tracked in #236.

PR #247, opened 2015-04-20
Changes to support routers with 4MB flash

sarumpaet:
Depends on https://github.com/freifunk-berlin/firmware-packages/pull/49 .

This is intended to make images for 4MB flash routers build again.
Space savings mostly are due to removing some non-essential packages around luci-mod-freifunk.
The postinst script hook allows to modify the firmware image just before the SquashFS image is built.
In this case, the postinst script for default_4MB removes some olsrd i18n files that are somewhat erroneously part of the luci olsrd packages. It could also be used to remove some of the profile_* config files, but all of these take only up about 4kB in the SquashFS image. The alternative to these one-liners would be to FORK ALL THE PACKAGES which I think would be far less transparent. For continued 4MB support we'll need this kind of patching sooner or later anyways.


sarumpaet:
Build 568 fine on a WR841v8. Overlay is 384kB, df reports 236kB used/148kB free, du /overlay reports 38kB used. Playing with dd if=/dev/urandom gives 100-110kB usable free space. Since JFFS2 compresses, this should be more for configuration data.
Redirecting to the Assistent on first boot doesn't work without luci-mod-freifunk installed, but this is a topic for another PR probably.

PR #248, opened 2015-04-25
[patches] LuCI bootstrap theme CSS improvements for input fields

sarumpaet:
Input fields and dropdown boxes used gray text color/background
which is typically used to indicate inactive fields.

This fixes https://github.com/freifunk-berlin/firmware/issues/213


booo:
Merged by hand.

PR #251, opened 2015-05-02
packages: replace wpad-mini with wpad to support DFS

lynxis:
This will increase the size of the image but DFS support is required for
5ghz outdoor nodes.


sarumpaet:
Have you actually tested the change? Other than the wiki I have not found any reference concerning DFS vs. wpad-mini; it's merely pointed out that wpad-mini lacks some auth variants.

lynxis:
yes. DFS only work when wpad is installed. However there are some bugs with DFS. See https://dev.openwrt.org/ticket/18656

Issue #123, opened 2014-10-16
use dns for vpn03 server instead of ips?


booo:
The current [openvpn configuration](https://github.com/freifunk-berlin/packages-berlin/blob/master/utils/freifunk-berlin-openvpn-files/uci-defaults/device-defaults#L27) uses ips instead of dns entries. Should we add one dns entry or multiple dns entries? I do think we should use dns entries instead of ips.

PR #264, opened 2015-07-17
Update to Chaos Calmer/add new targets

sarumpaet:
This updates to CC/OpenWrt trunk, adds support for the Nexx WT3020, the GL.iNet 6416A, and builds olsrd2.


booo:
I prefer to wait for a "stable" release of openwrt. Are there any reasons why we should experiment with the release candidate?

sarumpaet:
The idea is to base the next (0.2) kathleen release not on Barrier Breaker anymore but to update. As you can see in the commits, many features I worked on already depend on CC. We can update the OpenWrt base again later (to the actual release or whatever).

booo:
At the moment there is a automatic redirect from http to https. I think we should disable this otherwise we get a lot of confused users. @sarumpaet Did you look at this already?

booo:
For the https problem we have to change the `/etc/config/uhttpd` configuration file. The option we need to change: `option redirect_https '1'` Should be zero.

Should rpcd be part of the standard image? It is at the moment. Do we need it for something? If not I suggest we remove rpcd from the standard image and build it as module.

booo:
There is this strange new ip rule. Any idea where this rule comes from? There is no table 128 as far as I can see.
```
root@node01:~# ip rule
0:	from all lookup 128 
...
```

lynxis:
@sarumpaet please reword your commits. Don't use "see #123" because ticket might be change in future and referencing only to tickets is a bad idea. Do a full commit description to it.

lynxis:
CC released.

booo:
Ich hätte ja die ausstehenden Probleme gelöst und dann den Branche gemerged, aber gut...

booo:
Rpcd dependency was introduced by a commit in the luci repository: https://github.com/openwrt/luci/commit/199c8cbc4c32506ecfe89850615b88a3f0276dd3

PR #265, opened 2015-07-19
config.mk: update openwrt revision

booo:
Includes security updates for e.g. openssl. This should fix the broken openssl
in the current master too.

### DIFF
--- a/config.mk
+++ b/config.mk
@@ -3,5 +3,5 @@ SHELL:=$(shell which bash)
 TARGET=ar71xx
 PACKAGES_LIST_DEFAULT=default backbone
 OPENWRT_SRC=git://git.openwrt.org/14.07/openwrt.git
-OPENWRT_COMMIT=229d60fdb45c34902d402938e231c006f7c73931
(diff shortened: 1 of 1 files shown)


lynxis:
@booo please include a small Changelog (git log --oneline) into the commit message.

sarumpaet:
Hm, shouldn't we base the next (0.2) release on something newer than Barrier Breaker? See https://github.com/freifunk-berlin/firmware/pull/264 .

lynxis:
@sarumpaet I would like to do a release on 0.2. afterwards we change to chaos calmer.

sarumpaet:
What do you want to change for the 0.2 then?

booo:
I attached a short log to the commit body.

Maybe we should just push a minor release and upgrade to chaos calmer with a major release afterwards. I don't think it's a good idea to base a stable release on an unstable release candidate.

sarumpaet:
I don't quite see the point in making an extra BB-based release. Do we actually use OpenSSL anywhere? We use PolarSSL for OpenVPN and uhttpd as far as I can see. If you really want to make another BB release, please use 0.1.3. Merging in master/0.2.0 is of course also okay, but the CC merge will overwrite this anyways.

Issue #260, opened 2015-06-25
olsr topology view


booo:
For e.g. teaching purposes it would be nice to have a topology view for olsr in the router status interface. It's not necessary that the view is part of the default image. As far as I know there is a package for this somewhere in the openwrt ecosystem.

sarumpaet:
What kind of topology view? Obviously you are not referring to the build-in map view or the table topology view?

booo:
I talking about this: https://github.com/openwrt/luci/tree/master/applications/luci-app-olsr-viz

booo:
Done with the next package update so I'm closing this.

PR #281, opened 2015-09-18
[packages]: build luci-app-olsr-viz as package

booo:
Adds a package that can visualize the network topology.

Addresses: https://github.com/freifunk-berlin/firmware/issues/260


booo:
Ok. I'm merging this. @andrenarchy at the next freifunk class we should use this to give an impression of the mesh network.

PR #285, opened 2015-09-26
update revision of chaos calmer and routing feed

booo:
Untested update of chaos calmer release and routing feed.


booo:
I will update the feeds later this day.

PR #296, opened 2015-10-31
Revert: "[Makefile] describe version" (breaks filenames and migration)

sarumpaet:
Reverts freifunk-berlin/firmware#267

@lynxis , why did you merge this without further discussion? For master(-based) build this effectively sets the version to "15.05" now, breaking firmware filenames and migration. People who update to one of these firmwares will permanently break migration, see last lines of https://github.com/freifunk-berlin/firmware-packages/blob/master/utils/freifunk-berlin-migration/uci-defaults/999_freifunk-berlin-migration.sh .

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,9 +4,6 @@ include config.mk
 MAINTARGET=$(word 1, $(subst _, ,$(TARGET)))
 SUBTARGET=$(word 2, $(subst _, ,$(TARGET)))
 
-VERSION=$(shell git describe --tags --long --dirty --match 'v[0-9].[0-9].[0-9]*')
-REVISION=$(shell git log -1 --format=format:%h)
-
 # set dir and file names
 FW_DIR=$(shell pwd)
 OPENWRT_DIR=$(FW_DIR)/openwrt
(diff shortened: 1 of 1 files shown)


lynxis:
No one really cared. I merged it after some time.
I still prefer using tags, but neverthen less, if you still want to revert the version, I would like to use the revision part in our image.

sarumpaet:
Neither in the original pull request nor in its commit message you gave a reason why you fiddled with VERSION. Do give reasons for proposed changes in order to get comments.

I'll make another PR for removing VERSION (only). I think the current way of doing things is good - if you rely on git metadata for that, the tgz downloads that GitHub generates for tags will be broken.

We should include the *branch* name in buildbot downloads (in the path or in firmware filenames or somewhere else), but that's a different issue.

sarumpaet:
Actually. Where is REVISION visible in the firmware? I thought it's put in the motd, but %R there is the SVN revision.

sarumpaet:
Merging for now. We don't want more migrations broken.

lynxis:
@sarumpaet the revision is only included into images built by the imagebuilder, not in the /firmware-berlin/openwrt/bin/*

sarumpaet:
Thx, I'll have a look. For the record, /etc/banner substitutions are done in https://dev.openwrt.org/browser/trunk/include/version.mk?rev=47135#L75 .

lynxis:
The patch replaced line 9 (done over sed in Makefile).

PR #300, opened 2015-11-02
[configs] strip kernel exports

booo:
Maybe I'm just blind but I have the impression this does not change anything.

This should reduce the size of the kernel and the resulting firmware image. The
current configuration checks a build hash not just the kernel version so it's
nearly impossible to install other kernel modules than the modules that are
part of the release. This should not make the situation worse.

Change was proposed by Patrick Grimm.


sarumpaet:
We had this in https://github.com/freifunk-berlin/firmware/tree/features/strip_size_size a while ago as well. It does reduce kernel size a bit (the .lzma is 64kB smaller if I see it correctly in buildbot images). As far as I see the kernel hash (seems to be an OpenWrt invention) governs opkg dependencies only. You can always force install modules from other sources, and they might work, given the necessary symbols are there (which this patch removes), and that the modules were built with the same compiler and a compatible kernel (see CONFIG_MODVERSIONS). Isn't this one of the few reasons to stay with an OpenWrt release? ;-)

Still, as I think we should focus on a Berlin firmware, not some full OpenWrt replacement, fine for me. If anyone wants other kernel modules, create a PR or just switch to vanilla OpenWrt.

lynxis:
looks good to me

lynxis:
if we want to reduce the kernel size much more we could do:
* push devicetree support for atheros (upstream already got a lot of patches, it's name is ath79)
* remove boards from target/linux/ar71xx/config

PR #304, opened 2015-11-05
Set REVISION from Kathleen git (not OpenWrt); create README

sarumpaet:
See commit comments.
The README should help a lot when browsing buildbot directories (which so far had no indication of the git branch and revision that build was based on).


sarumpaet:
Somebody merge this? @lynxis ? The result can be seen here, for example: http://buildbot.berlin.freifunk.net/buildbot/unstable/ar71xx/702/README-kathleen-luci-mod-freifunk-ui.txt

booo:
I'm ok with the idea to create a file that somehow tells us which revision we are dealing with. I'm not happy with the current name especially for release branches.

Can we maybe rename the file to REVISION or something similar and put all the relevant build information there?

sarumpaet:
I changed the filename to VERSION-... so for a release it's "VERSION-kathleen-0.2.0.txt", for example.

booo:
I don't get why we need this file name suffix. Sorry ^^

sarumpaet:
Because "webserver sets mimetype based on extension" and "left click to view".
This is text/plain (and shown in browser directly):
http://buildbot.berlin.freifunk.net/buildbot/unstable/ar71xx/706/VERSION-kathleen-show_kathleen_revision.txt
This is application/binary (and quite annoying):
http://buildbot.berlin.freifunk.net/buildbot/unstable/ar71xx/706/default/md5sums

booo:
I suggest we reduce it to REVISION/VERSION.txt and maybe we can fix the mimetype for md5sums in the server configuration.

sarumpaet:
The purpose of including the branch name directly in the filename is to save people one click which is a nice touch I think. I don't see why leaving that away improves things...?

lynxis:
why not using git-describe based on tags? It would be much more unique and tracable.

sarumpaet:
a) The git commit hash is unique
b) The idea is to have the branch name (e.g., "fixing_openvpn") in the version file so that people looking through the buildbot directories can see immediately what branch that build is from.

Tags are relevant for releases only and are completely ignored on buildbot's side atm. If you want that to change, the order to do that is probably reaching an agreement first, then fix the buildbot, and the Makefile last.

SvenRoederer:
I think it's time to accept this RP as it is now. check the Version.txt of the current build including this stuff: http://buildbot.berlin.freifunk.net/buildbot/unstable/x86/600/VERSION.txt

it's proved to work and the infos provided seems handy to me.

SvenRoederer:
current build fails as of moved repo for olsr ... patch is in the pipeline (https://github.com/openwrt-routing/packages/pull/166)

PR #305, opened 2015-11-09
[patches] patch olsrd in routing feed to use procd as init process

booo:
This was proposed in a ticket here:

https://github.com/freifunk-berlin/firmware/issues/155

You can follow the upstream discussion here:

https://github.com/openwrt-routing/packages/pull/78

Initial test works for me. Please test too!

### DIFF
--- a/patches/001-feeds-routing-olsrd-procd.patch
+++ b/patches/001-feeds-routing-olsrd-procd.patch
@@ -1,0 +1,1858 @@
+Index: openwrt/feeds/routing/olsrd/Makefile
+===================================================================
+--- openwrt.orig/feeds/routing/olsrd/Makefile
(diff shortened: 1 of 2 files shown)


booo:
http://buildbot.berlin.freifunk.net/builders/ar71xx/builds/703

sarumpaet:
Just a side note (sort of out of scope of this PR): Both the original and the procd variant introduce a "hidden" iptables dependency, mess with configuration (by changing firewall settings) and are somewhat fundamentally broken: do a firewall restart while OLSRd/SGW is up and things stop working.

booo:
We should fix this upstream. There is a ticket in the upstream ticket system so I'm closing this one.

PR #315, opened 2015-12-25
Feature: common config

SvenRoederer:
as seen in #06da8aa it may prevent some work to separate between arch-specific and common config-option.
- so only one file has to be modified normally
- and the images are more equal


SvenRoederer:
this may help to keep the configurations of the differenct archs in sync

booo:
I will take a closer look at this as soon as I find some time.

SvenRoederer:
@booo as you merged PR https://github.com/freifunk-berlin/firmware/pull/317, did you had some time to look at this PR?
Getting this merged, will save some time, when a setting changes on all targets

booo:
I think you moved configuration from a specific target to common. E.g. the CONFIG_PACKAGE_om-watchdog=m configuration that should only be present in ar71xx configuration. Can you please double check. Apart from this minor problem I think it's fine to merge this.

SvenRoederer:
I'll show you some diffs between the original files and the result of the combined files tomorrow. Let's discuss this

PR #317, opened 2016-01-03
[configs] disable SSP_SUPPORT

booo:
Commit 35933b0bf8ad4f9d2ea3ed537df949af3ccd7258 tried to disable SSP_SUPPORT but
only removed the libssp package. The SSP_SUPPORT option enables the package
again and adds the ssp lib (libssp) to the image. By removing SSP_SUPPORT we
also remove the library.


booo:
I think this is a no-brainer so I merge this.

PR #322, opened 2016-01-11
test-build before final PR for upstreaming of improve version string

SvenRoederer:


SvenRoederer:
fails to apply as I had squashed a fix into 58dba13

PR #332, opened 2016-02-27
Improve layout / contens of Version.txt

SvenRoederer:
- move all repositories to the end of file
- one line per repository
- add OpenWRT-data

see http://buildbot.berlin.freifunk.net/buildbot/unstable/ramips/107/VERSION.txt


SvenRoederer:
@sarumpaet these are the recent changes, we spoke about

PR #333, opened 2016-03-03
Fix '/etc/init.d/olsrd6 restart' breaks IPv4 SmartGateway.

stspdotname:
Restarting olsrd for IPv6 clears SmartGateway firewall rules for both IPv4
and IPv6, and does not restore the IPv4 rules. As a result, IPv4 SmartGateway
functionality is broken after /etc/init.d/olsrd6 restart.
Add a patch to the olsrd package to fix this problem.

### DIFF
--- a/patches/023-olsrd-smartgw-rules.patch
+++ b/patches/023-olsrd-smartgw-rules.patch
@@ -1,0 +1,35 @@
+Index: openwrt/feeds/routing/olsrd/files/olsrd.init
+===================================================================
+--- openwrt.orig/feeds/routing/olsrd/files/olsrd.init	2016-03-03 00:06:12.447611233 +0100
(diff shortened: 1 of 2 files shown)


booo:
Can you try to fix this upstream too?

https://github.com/openwrt-routing/packages

Thanks for pointing this out and proposing a fix!

stspdotname:
Filed a pull request with upstream: https://github.com/openwrt-routing/packages/pull/155

booo:
I suggest that we update the routing feed instead of merging this. The commit was accepted upstream.

stspdotname:
Agreed. See https://github.com/freifunk-berlin/firmware/pull/336

booo:
Super seeded by #336

PR #335, opened 2016-03-06
Add profile for TP-Link TL-WA801N/ND routers.

stspdotname:
backbone_4mb image tested with TP-Link TL-WA801ND v2.1.
(new node clarke.olsr in Berlin mesh)


booo:
Looks good. Thanks for the contribution!

Issue #299, opened 2015-11-02
check CONFIG_CLEAN_IPKG build option


booo:
@sarumpaet My gut tells me that we implemented something like this somewhere.

"This removes all ipkg/opkg status data files from the target directory before building the root filesystem."

Maybe we should use this option.

sarumpaet:
We get great issues like #262 then and break opkg completely, just to save a few kB. This is a last resort thing at best.

booo:
Thanks for the clarification!

Issue #262, opened 2015-07-02
Table 113 in /etc/iproute2/rt_tables vanishes occasionally on sysupgrade


sarumpaet:
I have seen this on several sysupgrade processes now (from 0.1.0 to 0.1.1 and from 0.1.0 to 0.1.2, I think), but it doesn't happen always. Tables are set in /etc/uci-defaults/freifunk-berlin-olsrd-defaults, and the process for table 113 is the same as for the others. Strange. I'm tempted to add a "sync" to that file...

This breaks the SmartGateway functionality, but SGW still advertises the link.

sarumpaet:
Is this possibly a concurrency issue? /etc/init.d/freifunk-policyrouting edits that file as well...

sarumpaet:
Actually most likely I was upgrading from installations that had their pkg information stripped (old 4MB patch). Then, "opkg list-changed-conffiles" does not work, resulting in rt_tables not getting restored after the upgrade. Since freifunk-berlin-olsrd-defaults does not get past its guard after upgrades, it doesn't re-add olsr-tunnel then.

PR #341, opened 2016-03-22
more space on 4MB flash

SvenRoederer:
remove all opkg-related stuff from these small devices in order to gain some memory
-  this brings something around 120kB 
- and it's veeeery unlikely that the owner of a 4MB-router will do anything more than just default Freifunk

### DIFF
--- a/packages/default_4MB.sh
+++ b/packages/default_4MB.sh
@@ -9,6 +9,13 @@ rm -vf usr/lib/lua/luci/i18n/olsr.*
 echo "deleting opkg status-files ..."
 rm -rf usr/lib/opkg
 rm -rf etc/opkg*
+# see https://github.com/freifunk-berlin/firmware/pull/341 &
+#  https://github.com/freifunk-berlin/firmware/issues/262
+cat > lib/upgrade/keep.d/freiunk-berlin_no-opkg-info-on-4mb-workaround <<KEEPLIST
+/etc/iproute2/rt_tables
+/etc/firewall.user
+/etc/vnstat.conf
+KEEPLIST
 # as this will be included into image for some reason, even it's
 # not listed for inclusion
 echo "manually removing usign ..."
--- a/packages/default_4MB.sh
+++ b/packages/default_4MB.sh
@@ -6,3 +6,11 @@
 
 echo "Deleting OLSR i18n files..."
 rm -vf usr/lib/lua/luci/i18n/olsr.*
+echo "deleting opkg status-files ..."
+rm -rf usr/lib/opkg
+rm -rf etc/opkg*
+# as this will be included into image for some reason, even it's
+# not listed for inclusion
+echo "manually removing usign ..."
+rm usr/bin/usign
+rm usr/bin/signify
--- a/packages/default_4MB.txt
+++ b/packages/default_4MB.txt
@@ -7,6 +7,8 @@
 -ppp
 -ppp-mod-pppoe
 -wpad
+-opkg
+-usign
 
 # Defaults
 freifunk-berlin-dhcp-defaults
--- a/patches/016-imagebuilder-custom-postinst-script.patch
+++ b/patches/016-imagebuilder-custom-postinst-script.patch
@@ -3,12 +3,12 @@ Index: firmware/openwrt/target/imagebuilder/files/Makefile
 --- openwrt.orig/target/imagebuilder/files/Makefile	2015-04-07 00:27:34.000000000 +0200
 +++ openwrt/target/imagebuilder/files/Makefile	2015-04-09 01:28:05.000000000 +0200
 @@ -111,6 +111,9 @@
- ifneq ($(USER_FILES),)
  	$(MAKE) copy_files
  endif
+ 	$(MAKE) package_postinst
 +ifneq ($(CUSTOM_POSTINST_SCRIPT),)
 +	(cd "$(TARGET_DIR)" ; bash $(CUSTOM_POSTINST_SCRIPT))
 +endif
- 	$(MAKE) package_postinst
  	$(MAKE) build_image
  
+ package_index: FORCE


sarumpaet:
I think there's an OpenWrt build option for this (CONFIG_CLEAN_IPKG, see #299), plus doing this has weird side effects that might manifest only on subsequent upgrades (see #262). Still, on 4MB routers it's a good idea; for #262 we could add the ip_routes file to keep.d.

SvenRoederer:
yes, removing opkg per default (CONFIG_CLEAN_IPKG) seems also dangerous to me, as there is no way to use opkg anymore on these routers.
initial I "played the OPKG-card" as the image become so big, that there was no room for changing the password as of "no space on disc".

regarding upgrades to newer versions, I assume that all relevant settings are handeled by uci-default scripts as noe. And these getting setup during buildtime of image as normal and as used on the "large" flash units.

sarumpaet:
As explained in #262, rt_tables gets setup properly on initial setup by uci-default scripts, but the upgrade process resets the rt_tables file (normally, /usr/lib/opkg/info/ip.conffiles would guard it, but this patch deletes that), and uci-defaults isn't run again because of its guard, so after sysupgrade we end up with rt_tables missing table 113, and thus broken SmartGateway.

Either we move setting up table 113 before the guard in uci-defaults or put rt_tables in some keep.d so that it gets preserved on upgrades. I think the latter is better.

I wouldn't be surprised if there were other similarly hidden problems introduced by deleting opkg info though.

SvenRoederer:
hmm, still wondering of the guard()-thing .... but now have kind of an idea ...

what about spliting the current uci-default up in 2 files? one for olsr-routing, without guard, an other for uci-settings (olsr-defaults) with guard().

sarumpaet:
That would introduce a new configuration path that is difficult to explain, doesn't really make sense on >4MB devices, and will introduce subtle differences between devices (users adding stuff manually to rt_tables will see it working on 8MB devices and breaking on 4MB devices).

Why not fixing the actual thing we break either directly (adding a keep.d file in the olsr-defaults package is two lines in its Makefile) or fix it where we break it (`echo "/etc/iproute2/rt_tables" > /lib/upgrade/keep.d/olsr-defaults` in packages/default_4MB.sh).

glycoknob:
sorry for hijacking the pr - we in weimar have also space problems - If you need more space and only want to target a specific device you could also disable drivers not used for the router in the kernel config - i've tried this with a 841n  - https://gist.github.com/glycoknob/4026568b3de3f96780fb - the image will only work on the specific device through.

sarumpaet:
@glycoknob Thanks, we played with this a while ago, but back then the relatively little space gains seemed not worth the added effort of juggling another kernel config. How much more space do the changes you link to gain?

@SvenRoederer Concerning sysupgrade, I compared `cat /lib/upgrade/keep.d/*|sort` and `cat /usr/lib/opkg/info/*.conffiles|sort`, seems we should keep.d these files:
* /etc/iproute2/rt_tables
* /etc/dropbear/dropbear_*
* /etc/firewall.user
* /etc/opkg.conf
* /etc/sysupgrade.conf
* /etc/vnstat.conf

...otherwise, all these files will be set to defaults on sysupgrade if opkg info isn't present. So the echo to keep.d in default_4MB.sh seems to be the sensible approach (as it keeps the workarounds to another workaround in one place instead of scattering them all over the place). Easiest might be to just do `cat /usr/lib/opkg/info/*.conffiles|sort > /lib/upgrade/keep.d/no-opkg-info-on-4mb-workaround` in packages/default_4MB.sh before the rm.

SvenRoederer:
after a short trip into sysupgrade and "opkg list-changed-conffiles" I found the point sarumpaet is worrying on. I was not aware, that changed files marked as conffiles for opkg will be preserved automatically by sysupgrade.

So what about "cat /usr/lib/opkg/info/*.conffiles | grep -v /etc/config" to generate the keep.d-file?
This will of course include all known "conffiles", but the files are at least
- consistent between >4MB units and the reduced size-units
- etc/passwd and friends are included, so the password is also preserved (I did not realize the unset password on the recent tests with my approach)
- sshd-keys are preserved

Not sure on the side-effects of the unchanged files preserved by this approch, but it seems acceptable to keep them up-to-date by the migration-package

For reference a list of conffiles from a normal-build (TPL WR1043)
> /etc/hosts
/etc/inittab
/etc/group
/etc/passwd
/etc/shadow
/etc/profile
/etc/protocols
/etc/services
/etc/shells
/etc/sysctl.conf
/etc/rc.local
/etc/sysupgrade.conf
/etc/collectd.conf
/etc/dnsmasq.conf
/etc/dropbear/dropbear_rsa_host_key
/etc/dropbear/dropbear_dss_host_key
/etc/firewall.user
/etc/iproute2/rt_tables
/etc/opkg.conf
/etc/opkg/customfeeds.conf
/etc/vnstat.conf

SvenRoederer:
@sarumpaet seems that you posted your previous comment during I was writing mine.
But the final outcome seems to be very the same. so your minimal list is just missing "etc/passwd and friends".

sarumpaet:
@SvenRoederer See my previous comment. /etc/passwd is already covered by /lib/upgrade/keep.d/base-files-essential, but adding it to another keep.d file won't hurt.

glycoknob:
@sarumpaet: Tested it again, it's about 50kb gain for openwrt-ar71xx-generic-vmlinux.lzma no difference on the final image due to squashfs - you are right it's probably not worth the hassle.

SvenRoederer:
on a regular image I just compared the list of conffiles for opkg and the list generated by keep.d. Found the following files difference:
> /etc/dnsmasq.conf
/etc/firewall.user
/etc/iproute2/rt_tables
/etc/protocols
/etc/services
/etc/sysupgrade.conf

"protocols" & "service" I'll ignore, as nomally nobody will change here something I'll ignore these files. And user can still add them to sysupgrade.conf manually if he needs them preserved.
Same for "dnsmasq.conf"

I feel, preserving unchanged conffiles has the risk of keeping "legacy" configs which we have to migrate manually and to invent something again, what OpenWRT would normally do for us

sarumpaet:
You're right! :)

SvenRoederer:
@sarumpaet , @booo can you check on the idea I had yesterday evening --> https://github.com/SvenRoederer/freifunk-berlin-firmware/blob/wip/remove_opkg_on_4MB-flash/packages/default_4MB.sh

this can replace the hack we just merged by something which acts like "opkg list-changed-conffiles". it collects all conffiles, known to opkg and copies these hashes to the file (during buildtime). on runtime it just compares the known hashes with the stores one and adds them to the list of files to backup.
Runtime is triggered by sysupgrade-hook.

sarumpaet:
@SvenRoederer The sysupgrade-hook is a good idea! Shorter than storing and comparing sha256s is
```if ! cmp -s "/overlay/$i" "/rom/$i"; then echo "$i was modified locally!"; fi; done;```
I'd also write one sentence describing the issue on top of the code (additionally to the GitHub links) as people tend to not follow links. ;-)
I can do both things if you are busy. Let me know.

SvenRoederer:
I like this idea of comparing to the "ROM"-files. Even the output will be different from "opkg list-changed-conffiles", but as we are only interested in files differing from the Kathleen-defaults this should be fine.
Only having the opkg-conffiles suffixed to the script makes it more readable too.

Let me push some comments I recently added, then I'm happy if you continue to work on this code.

SvenRoederer:
@sarumpaet you can find my recent code here: https://github.com/SvenRoederer/freifunk-berlin-firmware/tree/send_up/remove_opkg_on_4MB

lynxis:
Hey, I talked with @bittorf from freifunk weimar about 4mb routers.

- replace with opkg binary with wget download & execute https://github.com/bittorf/kalua/blob/master/openwrt-addons/etc/kalua/software#L374
- remove ALL kernel debug messages https://github.com/bittorf/kalua/blob/master/openwrt-build/build.sh#L2320
- remove 11s https://github.com/bittorf/kalua/blob/master/openwrt-build/build.sh#L2320
- remove wlan debug

PR #348, opened 2016-04-26
remove opkg on 4MB devices

SvenRoederer:
with current build there is not enough space on jffs, which fails some boot-scripts (e.g. no generation of SSH-keys)

before:
> Firmware Berlin (kathleen 0.2.0-alpha+518d4fa rev 518d4fa)
root@OpenWrt:/# df -h
Filesystem                Size      Used Available Use% Mounted on
rootfs                  320.0K    260.0K     60.0K  81% /

after:
>  Firmware Berlin (kathleen 0.2.0-alpha+c4fa160 rev c4fa160)
root@OpenWrt:/# df -h
Filesystem                Size      Used Available Use% Mounted on
rootfs                  384.0K    244.0K    140.0K  64% /

### DIFF
--- a/packages/default_4MB.sh
+++ b/packages/default_4MB.sh
@@ -9,6 +9,13 @@ rm -vf usr/lib/lua/luci/i18n/olsr.*
 echo "deleting opkg status-files ..."
 rm -rf usr/lib/opkg
 rm -rf etc/opkg*
+# see https://github.com/freifunk-berlin/firmware/pull/341 &
+#  https://github.com/freifunk-berlin/firmware/issues/262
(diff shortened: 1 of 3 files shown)


sarumpaet:
Thanks a lot! But instead of the static keep.d file we should gather from the conffiles list as described in https://github.com/freifunk-berlin/firmware/pull/341#issuecomment-204783577 and your following comment.

SvenRoederer:
@sarumpaet can You implement this, as I'm short in time till the weekend

SvenRoederer:
I'll close this, as this is the same as #341 (beside that this one was rebased manually)

PR #349, opened 2016-04-26
ar71xx: add GL-AR150 support

buch8:
Add support by adding the patch for CC by GL to the patches directory and add new target.


SvenRoederer:
Hi looks great so far, but 
- can you rebase both commits into one
- fix indent in "target/linux/ar71xx/base-files/lib/ar71xx.sh"
- fix indent in "target/linux/ar71xx/patches-3.18/911-MIPS-ath79-add-gl_ar150.patch"
- fix indent in "arch/mips/ath79/machtypes.h"

are you testing here and later planning to upstream to OpenWRT?

SvenRoederer:
@NeoRaider this may also interest the Gluon community

buch8:
Better now?

GL said they will try to do a backport to CC. I wanted to have this PR in place in case kathleen 0.2.0 is released before the backport is accepted.

SvenRoederer:
merged and compiles locally.
please test on real Hardware. The current master is good for kathleen-0.2, for Devices with more than 4MB flash

buch8:
Tested with build 903. Looks good working as a mesh node.

SvenRoederer:
Great!!!
As GL will try to upstream, we'll keep this patch till then.

buch8:
Build 980 works with AR150.

MT300N or MT300A, use a mediatek chip, which doesn’t support simultaneous SSID broadcasts.

SvenRoederer:
thanks for checking.

but why no multi-sid on MT7620A? Hardware seems to support this as per DS (https://www.anz.ru/files/mediatek/MT7620_Datasheet.pdf) - 16 Multiple BSSID.

buch8:
> but why no multi-sid on MT7620A?
It was mentioned in the GL forum. Maybe it's only a problem with simultaneous mesh? It was in the context of mesh use.

NeoRaider:
Mediatek hardware (like most non-Atheros chips) doesn't supprt AP+IBSS. That's the reason we also support 11s to replace IBSS in Gluon now.

sarumpaet:
@NeoRaider We have a patch for Mediatek/rt2x00 AP+IBSS, see https://github.com/freifunk-berlin/firmware/blob/master/patches/001-rt2x00_allow_adhoc_and_ap.patch . BTW Atheros doesn't support AP+IBSS in vanilla Linux either, it's just that OpenWrt has a patch for this long since: https://github.com/openwrt/openwrt/blob/master/package/kernel/mac80211/patches/410-ath9k_allow_adhoc_and_ap.patch

NeoRaider:
@sarumpaet, I'm very aware that ath9k is hacked in OpenWrt to support this (and in Gluon, I've added a similar patch to ath10k).

I don't know about rt2x00, but for most drivers, there's actually a reason AP+IBSS is not supported (namely, hardware limitations). For newer (mt76xx) Mediatek chips I know for sure that they can't support this.

In the long term, we'll drop IBSS support in Gluon, so I don't really care about this too much...

sarumpaet:
@NeoRaider We have quite a few Nexx WT3020 (MT7620n) confirmed working. But sure, 802.11s is the way to go.

PR #352, opened 2016-05-05
remove opkg on 4MB devices - improved

SvenRoederer:
this is a improved version of the 4MB-workaround to remove opkg (https://github.com/freifunk-berlin/firmware/pull/341)

- tested it on a 4MB unit
- please check and merge


sarumpaet:
I don't have my 841 at hand but looks good.

PR #354, opened 2016-05-10
let Makefile accept more ENV-settings

SvenRoederer:
this is an attempt to get https://github.com/freifunk-berlin/buildbot/issues/33 solved.

The changes are completely hacky, and need some rework / cleanup / rebase. But with this code we can work on getting the buildbot reconfigured. See https://github.com/freifunk-berlin/buildbot/issues/33


SvenRoederer:
checked for correct building on #924 ... #926 for ar71xx

packages/backbone.txt
```diff
@@ -39,6 +37,20 @@ luci-app-owm
 luci-app-owm-ant
 luci-app-owm-cmd
 luci-app-owm-gui
+luci-theme-bootstrap
```
booo: This is part of another problem.

SvenRoederer: correct, this falls under "cleanup" I mentioned

booo:
@SvenRoederer can you cleanup the branch for further inspection?

booo:
How do I start a proper build after this change? Do I have to specify a BUILDTYPE? At the moment I can build without specifying a BUILDTYPE. Is this right?

SvenRoederer:
sane defaults are stored in "config.mk". final version will build "unstable" flavor and REPO_PATH as per openwrt/.config (taken from configs/common.config)

SvenRoederer:
@booo please review and comment

config.mk
```diff
@@ -4,4 +4,7 @@ TARGET=ar71xx
 PACKAGES_LIST_DEFAULT=default backbone
 OPENWRT_SRC=git://git.openwrt.org/15.05/openwrt.git
 OPENWRT_COMMIT=64e116779c0f7da6d98068b8e7c50f528c8a91f2
+REPOPATH=http://buildbot.berlin.freifunk.net/buildbot/unstable/%T/928/packages/
```
lynxis: What's the 928?

lynxis:
@SvenRoederer why not replacing the variable assignment of config.mk with ?= and overwrite these when running make?
```
OPENWRT_COMMIT?=64e116779c0f7da6d98068b8e7c50f528c8a91f2.
```
The variable will only set if not already defined. This way you can set the variable when calling make.

Makefile
```diff
@@ -128,7 +131,11 @@ firmwares: stamp-clean-firmwares .stamp-firmwares
 	rm -rf $(IB_BUILD_DIR)
 	mkdir -p $(IB_BUILD_DIR)
 	$(eval TOOLCHAIN_PATH := $(shell printf "%s:" $(OPENWRT_DIR)/staging_dir/toolchain-*/bin))
+ifeq ($(BUILDTYPE),unstable)
```
lynxis: A clean build should only contain one single file matching the regex OpenWrt-ImageBuilder-*.tar.bz2

Makefile
```diff
@@ -192,7 +199,11 @@ firmwares: stamp-clean-firmwares .stamp-firmwares
 	# copy imagebuilder, sdk and toolchain (if existing)
 	# remove old versions
 	rm -f $(FW_TARGET_DIR)/OpenWrt-*.tar.bz2
+ifeq ($(BUILDTYPE),unstable)
 	cp -a $(OPENWRT_DIR)/bin/$(MAINTARGET)/OpenWrt-*+$(FW_REVISION)*.tar.bz2 $(FW_TARGET_DIR)/
```
lynxis: same here.

Makefile
```diff
@@ -46,6 +46,9 @@ openwrt-clean: stamp-clean-openwrt-cleaned .stamp-openwrt-cleaned
 	  rm -rf bin .config feeds.conf build_dir/target-* logs/
 	touch $@
 
+openwrt-clean-bin:
```
lynxis: openwrt-clean-bin should be moved into a seperate commit.

config.mk
SvenRoederer: This was just an corresponding buildnumber to check that the setting is correctly applied.
The whole line gets removed finally.

Makefile
SvenRoederer: please suggest a regex covering the normal filename and also the filename with the version-string included.

config.mk
```diff
@@ -5,3 +5,5 @@ PACKAGES_LIST_DEFAULT=default backbone
 OPENWRT_SRC=git://git.openwrt.org/15.05/openwrt.git
 OPENWRT_COMMIT=64e116779c0f7da6d98068b8e7c50f528c8a91f2
```
SvenRoederer: @lynxis changing the way of specifying the OpenWRT-commit to be used is not the intention of the issue

lynxis: @SvenRoederer you don't have to change this if you want to create a "special" build. let's say you create a new config.mk variable called REPO_PATH and override it via make ARGUMENT when building a release.

lynxis:
@SvenRoederer I much more like a more general approach which makes override via make arguments.
So put the logic for BUILD_TYPE into the buildbot configuration and override the values via make variable.

Taken from https://github.com/freifunk-berlin/buildbot/issues/33
 * Download-path of opkg-sources inside image **1)**
 * upload-path of build images to firmware-server (not sure how this is done atm) **2)**
 * not including the commit-hash in the filenames & OpenWRT-version **3)**

1) Isn't this already implemented by the buildbot?
https://github.com/freifunk-berlin/buildbot/commit/109c3b3833ddb570e6e1128ef2ef1eefcc25b637

2) looks like this is already implemented in the buildbot configuration.

3) how do you think those files shoud be named? can you give an example of the OpenWrt/LEDE variable involved in this and how the file should be named?

Makefile
SvenRoederer: it might be, that some changes in the Makefile may have made this not required anymore. But initially I had some trouble here with multiple files matching.
Needs to be checked with current code

lynxis:
@SvenRoederer can we close this PR now?

PR #360, opened 2016-05-29
Makefile: include common.config before $(TARGET).config

lynxis:
When including first the common.config the target config can override certain symbols
if needed. Make a target more flexible.


lynxis:
@SvenRoederer

PR #361, opened 2016-05-29
Makefile: ensure quilt push -a can't fail when already on top of the …

lynxis:
…patchset

`quilt push -a` fails if a patch is broken, but also fails when calling
it on a patch series which is already applied. For this reason we can
not ignore the return code of `quilt push -a`.
Calling `quilt pop -a` and ignore it's return code before calling
`quilt push -a` fix this problem.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -70,6 +70,9 @@ pre-patch: stamp-clean-pre-patch .stamp-pre-patch
 # patch openwrt working copy
 patch: stamp-clean-patched .stamp-patched
 .stamp-patched: .stamp-pre-patch
+	# quilt push -a might fail if already applied all patches
+	# call quilt pop -a to go to the bottom. ignore return code of quilt pop -a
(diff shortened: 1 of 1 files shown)


booo:
Maybe we should just introduce a `patch-clean` target and call this before applying all patches?

SvenRoederer:
I did some work on a "patch-clean" target already ... let me sort it out

lynxis:
@SvenRoederer any news here?

SvenRoederer:
pls have a look @ https://github.com/SvenRoederer/freifunk-berlin-firmware/tree/send_up/Makefile_add_unpatch

lynxis:
@SvenRoederer if you would use rebase instead of merge, the history would be readable.
The unpatch looks ok. Please create a PR.

lynxis:
superseeded by #384

PR #359, opened 2016-05-29
[RFC] add PATCHES.rst to introduce a counting structure

lynxis:
helps maintaining our patches when backported patch comes first

PR #363, opened 2016-06-05
created naming-scheme for patches

SvenRoederer:
as discussed in https://github.com/freifunk-berlin/firmware/pull/359


lynxis:
merged.

PR #364, opened 2016-06-05
append subtarget name to all platforms

lynxis:
When subtarget support was introduce, it was only used for target ar71xx_mikrotik
and none of the old targets e.g. ar71xx was renamed including it's subtarget.
Every platform has a subtarget, even when it's generic.
Rename those platform into $TARGET_$SUBTARGET
LEDE requires to know the subtarget, because it has a different bin/ directory
layout. It uses /bin/$TARGET/$SUBTARGET/foobar.sysupgrade.squashfs.bin

ar71xx -> ar71xx_generic
mpc85xx -> mpc85xx_generic
ramips -> ramip_mt7620
x86 -> x86_generic

Signed-off-by: Alexander Couzens lynxis@fe80.eu


lynxis:
https://github.com/freifunk-berlin/buildbot/pull/35

SvenRoederer:
I think we should focus on getting this release out and then starting to switch to LEDE / make buildsystem compatible to LEDE & OpenWRT

lynxis:
I agree with you about the release, but both patches doesn't do much, they address a missing refactoring when subtargets introduced.

SvenRoederer:
@lynxis please have a look at https://github.com/SvenRoederer/freifunk-berlin-firmware/commit/0c0fae88588c2770c01edf96b3efc36f51eee325
to avoid issues with "_" in subtraget-names on:
- adm5120
- lantiq
- x86
- xburst
I suppose that we move to "-" as separator between `MAINTARGET` and `SUBTARGET`

SvenRoederer:
This might become superseded by #412

SvenRoederer:
Merged with PR #412

PR #368, opened 2016-06-12
Makefile: OpenWRT-config should be updated after all is patched

SvenRoederer:
- update OpenWRT
- update and install feeds
- patch all
- then run OpenWRT defconfig

this will prevent a patch which changes a feed-config-paramter to fail to apply


lynxis:
merged into master.

PR #369, opened 2016-06-12
fix VERSION_REPO override by passing it to make

SvenRoederer:
This implements the way of setting the URL of the OpenWRT-package-repo as planned by https://github.com/freifunk-berlin/buildbot/pull/34.

This supersedes PR #354 and reflects comments:
- "openwrt-clean-bin should be moved into a seperate commit"
- "A clean build should only contain one single file matching the regex OpenWrt-ImageBuilder-*.tar.bz2" --> put this into a separate commit, which can be removed if this "hack" is not needed

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -27,6 +27,9 @@ FW_REVISION=$(shell $(REVISION))
 ifndef BUILDTYPE
 $(error BUILDTYPE is not set)
 endif
+ifeq ($(filter release unstable,$(BUILDTYPE)),)
+ $(error invalid BUILDTYPE "$(BUILDTYPE)")
+endif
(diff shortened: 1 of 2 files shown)


lynxis:
NACK
- I don't see why we need this feature? Just override the specific variable if you do for a release.
- It makes also the code more complex and bloated.

SvenRoederer:
Any proposal of a variable to change the resulting image-filename?
We can use this way also to change some CONFIG-settings different on "unstable" and "releases", e.g. no SDK or Toolchain on unstable builds.

lynxis:
works now.

PR #374, opened 2016-07-06
patches: backport fixes for build with gcc 6

booo:
With gcc 6.1.1 and archlinux the firmware does not build anymore without these
patches. Patches backported from openwrt trunk with minor changes.


lynxis:
merged into master.

PR #377, opened 2016-07-30
profiles: add TP-link MR3220 for ar71xx

SvenRoederer:
V1, V2 have 4MB Flash / 32MB (Ram)

as per Userrequest (https://lists.berlin.freifunk.net/pipermail/berlin/2016-July/033846.html)


SvenRoederer:
Do we want to add this profile to the default builds?

lynxis:
merge in master.

PR #383, opened 2016-08-07
Makefile: add target openwrt-clean-bin

SvenRoederer:
this cleans up the build OpenWRT-files before starting a new build. This
might prevent the freifunk-build-process to look for the correct binary-
file on further processing, as there ist only one file.

Since commit 34d519c individual filenames are created per build, which
ended up in having the same type of buildresult in multiple versions.

Please check if prereq for ".stamp-openwrt-cleaned" are correct


lynxis:
LGTM

SvenRoederer:
merged

Issue #178, opened 2014-11-16
Shorten image file name for TL-WR841N N v9


stubb:
Just tried to flash a TL-WR841N N v9 with the image file openwrt-ar71xx-generic-tl-wr841n-v9-squashfs-factory.bin. The process failed with an error that the filename is too long. I recommend to shorten it, which was the way to get it going for me.

basros:
This issue still exists!

Flashing http://buildbot.berlin.freifunk.net/buildbot/unstable/mpc85xx/568/backbone/openwrt-0.2.0-alpha-mpc85xx-generic-tl-wdr4900-v1-squashfs-factory.bin is impossible without shortening the firmware filename. 
Since our flashing howto only mentions TL-WR841N v8.4 but this issue most probably also applies to every other TP-Link, we should consider updating our documentation/howto or just change firmware filename creation.

stubb:
In my opinion using a shorter filename during creation is better because it avoids the error notification and the additional shortening step on the user side.

lgierth:
@dangowrt told me this is an issue with most TP-Link devices (or even all of them). OpenWrt started adding its version to the image filename with Chaos Calmer, which now fails the filename length check present in the TP-Link vendor firmware.

So the issue is not actually with TP-Link, but with OpenWrt deliberately breaking the exact purpose of the squashfs-factory images.

booo:
Are we happy with the proposed format? An example:

```
kathleen-0.2.0-alpha-cpe210-220-510-520-factory.bin
```

I wonder if the user is confused by the new format and how many howtos we have to change.

sarumpaet:
I'm happy with the format. ;)
We have to change URLs for each release anyways (on download page, wiki, config.berlin), so this change isn't extra work really. Even the Howto doesn't link directly.

stubb:
The new format has 52 chars in comparison to 56 of the initial format. Did someone checked the max supported length? If the new format is too long I propose to use an 'a' instead of the word 'alpha'.

sarumpaet:
i) The "alpha" will be omitted in releases anyways.
ii) You seem to compare the wrong filenames.
Old: openwrt-ar71xx-generic-tl-wr842n-v2-squashfs-factory.bin (57 chars)
New: kathleen-0.2.0-(alpha-)tl-wr842n-v2-factory.bin (40/46 chars)

stubb:
[...]

PR #412, opened 2017-01-04
make build infrastructure LEDE compatible

SvenRoederer:
this will prepare the way to switch over to LEDE


Makefile
```diff
@@ -121,33 +121,7 @@ firmwares: stamp-clean-firmwares .stamp-firmwares
...
-	      CUSTOM_POSTINST_PARAM="CUSTOM_POSTINST_SCRIPT=$$PACKAGES_PATH/$$PACKAGES_FILE.sh"; \
-	    fi; \
-	    PACKAGES_FILE_ABS="$$PACKAGES_PATH/$$PACKAGES_FILE.txt"; \
-	    PACKAGES_LIST=$$(grep -v '^\#' $$PACKAGES_FILE_ABS | tr -t '\n' ' '); \
-	    $(UMASK);\
-	    echo -e "\n *** Building Kathleen image file for profile \"$${PROFILE}\" with packages list \"$${PACKAGES_FILE}\".\n"; \
-	    $(MAKE) -C $(IB_BUILD_DIR)/imgbldr image PROFILE="$$PROFILE" PACKAGES="$$PACKAGES_LIST" BIN_DIR="$(IB_BUILD_DIR)/imgbldr/bin/$$PACKAGES_FILE" $$CUSTOM_POSTINST_PARAM || exit 1; \
-	  done; \
-	done
+	#mv $(IB_BUILD_DIR)/$(shell basename $(IB_FILE) .tar.bz2) $(IB_BUILD_DIR)/imgbldr
```
sarumpaet: Commenting out this but at the same time removing the explaining comment isn't very good style.

Also, it's not like shortening the path was added because we were just bored; it was necessary.
Have you tried building in, e.g., /home/user/project/freifunk/firmware? Previously, the build failed then (also on the buildbot, for some targets).

Makefile
```diff
@@ -166,25 +140,10 @@ firmwares: stamp-clean-firmwares .stamp-firmwares
...
-	  for FILE in $$TARGET_DIR/openwrt*; do \
-	    [ -e "$$FILE" ] || continue; \
-	    NEWNAME="$${FILE/openwrt-/kathleen-}"; \
-	    NEWNAME="$${NEWNAME/ar71xx-generic-/}"; \
-	    NEWNAME="$${NEWNAME/mpc85xx-generic-/}"; \
-	    NEWNAME="$${NEWNAME/squashfs-/}"; \
-	    mv "$$FILE" "$$NEWNAME"; \
-	  done; \
-	done;
+	./assemble_firmware.sh -p "$(PROFILES)" -i $(IB_FILE) -t $(FW_TARGET_DIR) -u "$(PACKAGES_LIST_DEFAULT)"
```
sarumpaet: This seems to silently remove the feature of shortening/renaming firmware images?

SvenRoederer: yes, some lines have been dropped by Lnyxis.
but when when buildbot is ready for the new targets, I have a patch ready to get the old behavior back

Makefile
SvenRoederer: got it, so lets test it.
most of this is taken from Lynxis lede-branch and as long as the buildbot accepts the new SUBTARGETS (https://github.com/freifunk-berlin/buildbot/pull/41) we can fix things easily.

Makefile
SvenRoederer: but https://github.com/freifunk-berlin/firmware/blob/2f7f760bb9c659ff83317570526f61a81cdb6ef6/Makefile#L135 shows that the buildslaves are not failing with this commented out (e.g https://buildbot.berlin.freifunk.net/buildbot/unstable/ar71xx/1152/)

booo: @SvenRoederer I suggest that you readd the missing features like shortening/renaming the firmware images and then we merge.

lynxis: @sarumpaet @booo  that hack was removed by intention, because there is no more reason, why we need to shorten the paths.

lynxis: unless you find a reason, why we need to do so.

sarumpaet: @lynxis I take it you haven't had a look at https://github.com/freifunk-berlin/firmware/commit/fa4aafa7e594833efd18d113ab24dda77cfe9918 and https://github.com/freifunk-berlin/firmware/issues/178 . `git blame` makes finding that really easy.

Aside from some routers having problems with filenames that long, we're not building _openwrt_ (or LEDE) images here, we're building _kathleen_ images. Also, look at the mess that http://buildbot.berlin.freifunk.net/buildbot/unstable/ar71xx-generic/2/default/ is currently. Speaking of that, http://buildbot.berlin.freifunk.net/buildbot/unstable/ is quite messy now, too. Sigh. I'd love if we could at some point work on the firmware, not on its build infrastructure.

Makefile
lynxis: looks more likely a leftover. Whoever did the originally shorting mv. I don't see any ! points in doing workarounds without even reporting the bug!!!!!!!! Not even talking about fixing it. Really please think for **long term**.

Makefile
lynxis: @sarumpaet It would be easy for anybody to understand why we need such strange workarounds, if there would be a good comment within the code. I can not remember, what I did when I written that commit, to much time pasted. But I'm looking up now would be:
- git blame
- see commit fa4aafa
- git show fa4aafa
- oh nice, there is a reference, then let's start webbrowser to see what's the original reason is
- #178 oh great. Just have to read 18 comments to get a overview, what's the original reason and why this commit might fix it.

**Not an easy way.**
Quite time consuming for horrible code, uncommented code.
Commit messages with "See url..." is **NOT** helpful. Please don't create indirections in commit messages, especially bug numbers, bug system, url tend to change when moving from one system to another. Describe what you do and why you do it. Especially when you introduce strange code and add a comment into the strange code too. Yes I am also tend to be lazy from time to time. Please poke me then.

But anyway, I can not find any upstream bug report about this problem? Also the first substitution can be done by configuration.
Removing the platform prefix "mpc85xx-generic-" seems to be a good feature for upstream as `make menuconfig` option.

Please as I stated in the other conversation. Think **long term**. It's important unless we will do the same work again, and again, and again, wasting time in porting patches and workarounds.

SvenRoederer: @bittorf mentions this upstream-thing in https://github.com/weimarnetz/weimarnetz/issues/216#issuecomment-164053487
Maybe he can tell more regarding upstream, even this might still be only relevant for OpenWrt and not discussed with the LEDE-guys

SvenRoederer:
please check this now and esp. d13c3d76c568ee6befb3885e8ef0a4634bc83b0a which is redoing the renaming.
It's not a one-liner, but feels more easy to read even for new ppl. get involved.

SvenRoederer:
Any feedback on this?
If no objections are brought up till the weekend I'd like to merge. This date is chosen in relation to the upcoming LEDE-17.01 branch (http://lists.infradead.org/pipermail/lede-dev/2017-January/005241.html). This branch is half the way to make the switch to LEDE.

booo:
I suggest that we wait till there is an actual release.

Last time I build LEDE mtr and iperf did not compile. Does this work now?

SvenRoederer:
Why waiting?
This commits only change some internals of the Makeporcess. The resulting firmware-images are still the same as used in Kathleen-0.2.0.
It just prepares the buildprocess to be compatible with OpenWrt and LEDE, so we can switch back and forth.

From what I've seen so far, there will come up some issues with updated LuCI anyway (with LEDE and OpenWRT-DD (in case this will ever released)). 

To not having a release-phase of 15 months again, we should be a bit more active and not wait longer and longer. This might end up in updating the. Software sme days before releasing, wich is not the best move.

Am 12. Januar 2017 17:42:35 MEZ, schrieb Philipp Borgers <notifications@github.com>:
>I suggest that we wait till there is an actual release.
>
>Last time I build LEDE mtr and iperf did not compile. Does this work
>now?

SvenRoederer:
regarding packages not building, have a look at branch https://github.com/freifunk-berlin/firmware/tree/SAm0815/build_with_lede.
From our default package-selection only "horst" had to be disabled, even this was in mid of Dec 2016

bobster-galore:
I think it's a nice idea to start the lede integration right now. As I understood, the builds are made from a well described point of lede builds and not the latest. We can check if there are still some things to solve. And I believe there will only marginal changes to lede 17.01 come up if so.

The parameter lede / openwrt for the build process can be chosen, weekly builds are openwrt. Things which can't be compiled will break and compiling continues with the next thingie. So no fatal images are produced automatically and the build process continues.

One main advantage is support for new devices e.g. EdgeRouter X without backports (wasted energy imho), regular reproducible freifunk builds would be really appreciated by MABB-Meko project.

Also we could get more familiar with cpe510 concerning transmission power in lede. There needs something 2b done to get satisfying transmission and may be also dfs.

At least we are on the straight way to Hedy lamar 0.1.0!

sarumpaet:
This PR needs a bit of `rebase --interactive`; of its 14 commits, many fix regressions introduced by other commits before. Such "junk" in the history makes `git blame` difficult to use, and defeats the purpose of the whole PR-based commit model. Let's rework/merge this at the meeting on wednesday?

SvenRoederer:
I was also thinking of rebase, but also liked the idea of keeping the original commits I cherry-picked from other branches. 
@lynxis @andrenarchy are you good with rebasing some of your commits and getting lost of the original authorship or having different content than you committed?

lynxis:
@SvenRoederer it's possible to preserve the authorship. do you need help with git to do so? I'm happy to help out.

booo:
I suggest that we meet on Wednesday and try to merge this. Once the merge is done we can discuss plans to switch to LEDE.

SvenRoederer:
It's very unlikely the I can manage to join

But hopefully https://github.com/freifunk-berlin/buildbot/pull/43 can /will be merged tomorrow anyway

SvenRoederer:
I'll have a look in the commit-messages and their order in the next hours

SvenRoederer:
btw: LEDE-17.01 branches have been created (http://lists.infradead.org/pipermail/lede-dev/2017-January/005404.html)

PR #417, opened 2017-01-15
embedd files directly into image, rather than patching

SvenRoederer:
This way we include the banner-file, and possible other files, directly and don't have to patch them.
So we can avoid some patching and pushing some changes to files is much more easy.

Verified to work in http://buildbot.berlin.freifunk.net/buildbot/unstable/mpc85xx/1092/ before squashing and rewording.


SvenRoederer:
I consider merging in the next days, as this will avoid additional patching when switching to LEDE / upgraded OpenWrt-file

booo:
Why don't we create symlinks to the toplevel directory of the files instead of coping the stuff around?

SvenRoederer:
just did as we do currently for th patches (https://github.com/freifunk-berlin/firmware/blob/master/Makefile#L53)

SvenRoederer:
forget about my last comment.

SvenRoederer:
Think my idea was, "use copy, as you later can modify the file in the build-tree, w/o altering the originals".
But ... ????

Let'S do the link-way, as we do for patches

SvenRoederer:
Hmm, I did not mention that I reworked the code correctly from copy to link?

SvenRoederer:
I suggest to reset the tree and give me some time for correcting

booo:
I suggest you either push a fix to the master branch or we revert the stuff. Force pushing to master is bad and github protects the master branch by a special hook as well.

As far as I can see there is a problem with the dependencies. Right?

SvenRoederer:
forcing on master is bad - true. But it's not impossible ...
esp with this uuuugly commit - faulty, introduces unneeded code, a complete non-matching commit-message. And as @lynxis mentioned it looks more nice to rebase such single commit into master than merging it an generation 2 commit-messages

booo:
I don't think a commit that fixes a mistake is that ugly. In the end mistakes happen and are part of the process.

But do as you please ;)

SvenRoederer:
finally merged with https://github.com/freifunk-berlin/firmware/commit/a66896d3ca02d1aff22be0c5b77e812ebd7596d4

Issue #418, opened 2017-01-19
issues seen with LEDE-base build


SvenRoederer:
this is just for collecting roughly the issues seen when building with LEDE-17.01-branch. Should be transformed into separate issues as soon as we switch to LEDE, or discard this issue when LEDE is not an option.

SvenRoederer:
* ff-assistent: requires setup of "private AP" to continue
* ff-assistent: shows textbox with OpenVPN-key paths inside
* default 4MB-images only build for TPL-WR701, other rootfs are to big
* buildbot "wg1337" fails, as of low disk-space

bobster-galore:
- on  cpe210 owm.lua doesn't work
- on  cpe210 olsrd isn't stable, forgets neighbours names or routes to them
- on  cpe210 olsrd doesn't know it's cpe210.olsr only ipv4-address works

bobster-galore:
- on cpe210 no data for hopglass & openwifimap while monitor is up to date

booo:
> on cpe210 no data for hopglass & openwifimap while monitor is up to date

If owm.lua does not work/report data hopglass & openwifimap does not have the data for visualization.

pmelange:
Based on branch SAm0815/build_with_lede:

The libraries referenced in /etc/config/olsr and /etc/config/olsr6 need to be updated.
olsrd_txtinfo.so.0.1 -> olsrd_txtinfo.so.1.1
olsrd_nameservice.so.0.3 -> olsrd_nameservice.so.0.4
olsrd_jsoninfo.so.0.0 -> olsrd_jsoninfo.so.1.1

The luci status page for configured interfaces does not show the actual data.
The luci status page for the smart gateway shows nothing although the plugin definately sets up a tunnel correctly.

Running a command such as 'echo "/gateways" | nc 127.0.0.1 9090' is not nicely indented and organized like it is in 0.2.0.  There is also some new information provided by this command.  Perhaps that is having in influence on the luci status pages.

pmelange:
Based on branch SAm0815/build_with_lede:

I disabled the 'berlin.freifunk.net' SSIDs on both 2.4Ghz and 5Ghz to prevent clients from attaching to my test node.  On the luci wireless status page, the SSID's shown for the disabled interfaces are not showing up as 'berlin.freifunk.net' but as 'intern-chXX.freifunk.net'.

SvenRoederer:
[...]

PR #420, opened 2017-01-31
switch to LEDE as base for next release

SvenRoederer:
- OpenWrt-development is nearly stuck 
- LEDE is going to release "LEDE-17.01"

As here are picked commits reaching back to may 2016 do not merge this, please rebase this onto master!


SvenRoederer:
@lynxis you are right.
This is just a PR of my working-base, as I'd like to collect feedback before investing time in cleaning up the commits.

So, I'll wait some more days for feedback and then start to rebase that this can be merged.

booo:
Please create a clean and minimal pull request for the switch to LEDE and then we start the review process.

I think it is consensus to switch to LEDE as soon as there is a stable release for LEDE.

SvenRoederer:
please check again, think ready to merge. More squashing I'll avoid.

slaves are building now:
* ar71xx-generic: build# 93
* ar71xx-mikrotik: build# 87
* mpc85xx-generic: build# 86
* ramips-mt7620: build# 80
* x86-generic: build# 88

booo:
What about the issues mentioned here: https://github.com/freifunk-berlin/firmware/issues/418

I think we should not enable telnet. The lede upstream configuration is fine to me.

The bbb-vpn is a non standard setup. We should stick to the mbedtls openvpn package as standard package for the standard firmware.

Do we have more information on the disk space problem?

SvenRoederer:
The intention of #418 is only to collect the issues for the time LEDE is not the underlaying base. So these issues need to be solved

Without telnet, it will be hard to connect to legacy nodes like switches or modems which are not ssh-enabled. telnet is also useful for network debugging.

OpenVPN: VPN03 also uses RSA1024.

4MB: see https://lede-project.org/supported_devices and https://forum.lede-project.org/t/should-lede-support-devices-with-only-4mb-flash/1018

SvenRoederer:
but most, if not all, the problems referenced in #418 we will also have with OpenWrt-DD, as this also uses the same feeds for packages, LuCI and Routing.
Only OpenVpn would work with current OpenWrt-master, but only as they still hang on the deprecated polarSSL

SvenRoederer:
As per LEDE-devel ML the release of 17.01. comes close.
Are there any NACKs to merge this till / at the "Firmware+Wizard Hackathon 18.2. 14:00 @ PH HQ" ?

SvenRoederer:
@booo:
- OpenVPN: use of the openssl-variant creates the most minimal issues with the currently rolled out structure. Certificates can be reused, user can switch to the new code w/o additional actions. the only point are the 4MB-devices, but there is a warning out long time already.
So this switch seems to me the most minimal diff regarding code and function.

@lynxis:
- px5g-embed-tls: fixed in recent force-push
- uhttpd-mod-tls: removed from config
- CONFIG_VERSION_CODE: I prefer to stay at the same style of code as we introduced in https://github.com/freifunk-berlin/firmware/blob/5a7fa79a1c4dc006cc5bc3c5cba8dd761a0ff1e4/Makefile#L86. Using the SED-style is probably the most minimal change. 
Even the SED-style is not the most beautiful code, but making it more nice I'd to leave for a commit out of the scope of the PR
- embedd files / banner: as it's in master, fixing this don't belong to this PR

lynxis:
let's merge it now.
There still might some problems with the version_code, but the old behaviour isn't great anyway so. So we can fix this later.

PR #426, opened 2017-02-25
added some buffalo ar71xx targets

hafu:
Added some build Targets:

- WZR-HP-AG300H / WZR-600DHP
- WZR-HP-G300NH2 / WZR-300HP
- WZR-HP-G450H

Especialy the WZR-HP-AG300H / WZR-600DHP is a common Hardware with dual Band WiFi, a lot of Memory and a fast CPU.


SvenRoederer:
thanks

PR #433, opened 2017-03-10
Makefile make tagets more atomic

SvenRoederer:
please only look at the changes in the Makefile

This moves the generation of the VERSION.txt into a separate target and splits up this ugly script-block.

if nobody sees an issue, I'll push only the Makefile-commit onto master.


booo:
Looks good to me. Maybe you/we want to do some more refactoring on the Makefile. Some steps are just a big mess.

SvenRoederer:
I'm working on it step-by-step ...

SvenRoederer:
merged in e7b8a36

PR #453, opened 2017-04-18
Improve makefile

SvenRoederer:
3 commits which
- testing for existing $TARGET-config (which was mainly useful when buildbot was building the old targets in parallel to the new naming (ar71xx vs. ar71xx-generic)
- make-target "lede-clean-bin" also removes the SDK and imagebuilder - please comment on this
- use separate commands to update feeds: is more easy to read and seems to have no practical difference


SvenRoederer:
the code itself is doing it's job since some time in my builds

PR #459, opened 2017-05-25
Build images from imagebuilder

SvenRoederer:
a commandline will look like:
make images IB_FILE=<file> TARGET=... PACKAGES_LIST_DEFAULT=fancy-image

and everyone can build a final image for the own needs. I use this to build specialized images for the UBNT-ERX.
The resulting image will also use the .ipks stored on buildbot without signing issues as a self compiled image would have.


SvenRoederer:
I know the check around ".stamp-images" is a bit uggly, but works as expected. Improvements are welcome.

Issue #414, opened 2017-01-08
save some disc-space when building on buildbot


SvenRoederer:
Think we can save some disc-space on buildslaves and makes buildprocess more nice by:
- building OpenWrt / LEDE
- copy all files we need to build our firmwarefiles (at least: imgbuilder, packages; in addition SDK)
- call ```make clean``` in upstream-firmware (this will remove ``´bin/``` and ```build_dir/target*```)

for better reading, we can also move the "Version.txt" section into a separate make-target

SvenRoederer:
the ```make clean``` we should per default only do on buildbot-builds, to keep the binaries when compiling locally.
PoC for detection is https://github.com/freifunk-berlin/buildbot/pull/42

SvenRoederer:
recent OpenWrt builds failing on slave wg1337, see https://buildbot.berlin.freifunk.net/builders/ar71xx-generic/builds/175

SvenRoederer:
still valid for LEDE 17.01 (https://buildbot.berlin.freifunk.net/builders/ar71xx-generic/builds/214)

bobster-galore:
Will there be a new behaviour? Gives moving files enough room for the builds?

SvenRoederer:
In 64dfaea the code of PR #452 was merged, which moves some files instead of copying them.
This saves probably 300 MB, but buildslave "wg1337" still fails when building target ar71xx-generic.

SvenRoederer:
What about removing ar71xx-generic from the list of valid targets for buildslave wg1337?

SvenRoederer:
https://github.com/freifunk-berlin/firmware/commits/issue414_v2 is an alternative approach to handle this.

SvenRoederer:
This was autoclosed, but still some working detection for the buildbot is missing. --> https://github.com/freifunk-berlin/buildbot/pull/42

SvenRoederer:
I pushed "Makefile: detect if we running on a special host" to branch next_release in order to merge into master.
This problem is known since >1year and there was only one comment, on this solution. It was that it's bad to have different rules / execution of the Makefile when running on Buildbot or having a local build. But this can be avoided by setting "SET_BUILDBOT" in config.mk (https://github.com/freifunk-berlin/firmware/commit/ec06e30ea452150e75b3e45ff23efa2a6d623200)

SvenRoederer:
so this code made it into master and Hedy-1.0.0

PR #460, opened 2017-05-29
Makefile: run imagebuilder in parallel

andrenarchy:
# What does it do?

Uses make's `-j` option to parallelize imagebuilder. The changes in this PR work as follows:

* first `IB_JOB_TARGETS` is initialized with make targets for each imagebuilder run, see `ib-generate-jobs.sh`
* instead of calling `assemble_firmwares.sh` make passes the jobs to `ib-generate-firmware.makefile` which automatically parallelizes the jobs according to the `-j` flag.
* `assemble_firmwares.sh` was renamed to `ib-generate-firmware.sh` and adapted to build only a single firmware

Note: each imagebuilder run runs in an isolated imagebuilder directory so there are no race conditions (thus consuming more disk space though depending on `-j`).

# Benchmark

Timings for ar71xx-generic `make` on machine with 16 cores:

* build [265](https://buildbot.berlin.freifunk.net/builders/ar71xx-generic/builds/265) (master) takes 40 minutes
* build [267](https://buildbot.berlin.freifunk.net/builders/ar71xx-generic/builds/267) (this PR) takes 24 minutes


andrenarchy:
Also note that there is an issue in imagebuilder with `BIN_DIR` that breaks some profiles (e.g., DIR505, also happens in master right now but it just seems to ignore the error instead of failing, see https://bugs.lede-project.org/index.php?do=details&task_id=590)

SvenRoederer:
I assume this change will extract the Imagebuilder for every ib-job separately. In this case this will make issue https://github.com/freifunk-berlin/firmware/issues/414 even more relevant (esp. fir Buildslave wg1337)

booo:
I suggest that we add a bit more documentation to the code, makefile, README and commit body.

Why don't we fix this upstream?

andrenarchy:
Imagebuilder is designed to build one profile with one package list. Adding parallelization there would require quite some rewriting there and break the way parameters are passed. We could try to improve imagebuilder such that parallel runs in the same directory don't lead to conflicts.

I'm happy to add some documentation. :)

booo:
> [...] break the way parameters are passed. We could try to improve imagebuilder such that parallel runs in the same directory don't lead to conflicts.

I don't understand that. Can you explain why parallel runs create conflicts? How does it break the way parameters are passed.

I agree it's work but maybe it's worth the effort.

andrenarchy:
If we have parallel imagebuilder `make` runs they may create/remove files in the same places. Not sure I want to dive deep into the imagebuilder code to check this. Maybe @lynxis has more knowledge here?

booo:
I patched my local LEDE based imagebuilder to accept multiple TARGET_DEVICES/PROFILEs. With this patch I can build all ar71xx devices in ~ 12s on my laptop. Calling the imagebuilder again and again for each profile causes useless fetches of the package index, download of packages and checksum calculation. We can circumvent these problems if we work on a patch that adds support for multiple TARGET_DEVICES/PROFILEs to the imagebuilder.

andrenarchy:
@booo  :+1: 
@lynxis do you see any problems getting this upstream?

booo:
I will try to come up with a proper patch that we can push upstream. The current implementation is a bit hacky.

andrenarchy:
I know. :)

booo:
The imagebuilder calls the board Makefile here:

https://github.com/lede-project/source/blob/master/target/imagebuilder/files/Makefile#L177

E.g. the Makefile for ar71xx:

https://github.com/lede-project/source/blob/master/target/linux/ar71xx/image/Makefile

If the target is e.g. generic we include for example the tp-link.mk:

https://github.com/lede-project/source/blob/master/target/linux/ar71xx/image/Makefile#L16

The tp-link Makefile adds a lot of TARGET_DEVICES to the namespace:

https://github.com/lede-project/source/blob/master/target/linux/ar71xx/image/tp-link.mk#L1027

https://github.com/lede-project/source/blob/master/include/image.mk

The board Makefile from here:

https://github.com/lede-project/source/blob/master/target/linux/ar71xx/image/Makefile#L64

calls BuildImage from the included image.mk file:

https://github.com/lede-project/source/blob/master/include/image.mk#L518

And somewhere in the BuildImage function we iterate through all the devices and build an image for each device:

https://github.com/lede-project/source/blob/master/include/image.mk#L561

If we add an option to overwrite the TARGET_DEVICES and LEGACY_DEVICES variables in the call chain we reduce the number of devices we build and speed up the build process.

booo:
Closing this due to inactivity. Nevertheless thanks for your time and work.

I still think we should improve the image builder upstream so everybody benefits from an improved build process.

PR #471, opened 2017-07-03
configs: build kmod-nf-nathelper-extra as module

booo:
The nf-nathelper-extra kernel module enables proper network address
translation for protocols such as pptp, gre, sip.

Package description:

https://lede-project.org/packages/pkgdata/kmod-nf-nathelper-extra

### DIFF
--- a/configs/common.config
+++ b/configs/common.config
@@ -82,6 +82,7 @@ CONFIG_PACKAGE_kmod-lib-crc-itu-t=m
 CONFIG_PACKAGE_kmod-lib-crc16=m
 CONFIG_PACKAGE_kmod-lib-crc32c=m
 CONFIG_PACKAGE_kmod-lib-textsearch=m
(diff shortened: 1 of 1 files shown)


SvenRoederer:
NAT is evil and bring problems!!!!

As we see here ....

Are these modules also loaded on the VPN-gateways?

booo:
Thanks a lot for this very helpful comment ;)

We should discuss if we want to integrate the modules by default in the image.

The modules are not loaded on the VPN-gateways as far as I can see. Maybe we should do this too.

SvenRoederer:
As we don't do NAT by default in the firmware, it doesn't make sense to include "nat-helper" by default.

booo:
In case we enable NAT on the gateway nodes by default we should consider adding the nat-helper.

Anyway I don't see a reason why we should not merge this specific change. There are multiple gateways in the network which would benefit from a available package for nat-helper.

PR #477, opened 2017-07-24
backport Ubiquiti ERX SFP to LEDE 17.01

SvenRoederer:
this patch adds a "cheap" backport of the Ubiquiti ERX-SFT support from LEDE-master to LEDE-17.01. So this patch adds the board "ubnt-erx-sfp" for ramips-mt7621 target.

I call it cheap, because it's not using common parts in a included DTS, just a copy of the ERX-dts modded for ERX-SFP.
also the GPIOs are not working as in Master.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -79,6 +79,7 @@ pre-patch: stamp-clean-pre-patch .stamp-pre-patch
 patch: stamp-clean-patched .stamp-patched
 .stamp-patched: .stamp-pre-patch
 	cd $(LEDE_DIR); quilt push -a
+	rm -rf $(LEDE_DIR)/tmp
 	touch $@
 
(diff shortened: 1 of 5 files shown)


SvenRoederer:
it'S basically to have theis board avail, that later upgrades, to a more advanced 17.01-version or next LEDE-release, will not require serial access to the router.

A bit of clean up I'll do in the commits, before somebody likes to merge

booo:
Do you think it's possible that we backport this upstream?

SvenRoederer:
follow this thread: http://lists.infradead.org/pipermail/lede-dev/2017-July/008394.html

When we remove the redundant code in Makefile and DTS, the chanches are good, but not in the state as it's now.
And the GPIOs should also work before sending upstream.

PR #495, opened 2017-12-10
in preinit and failsafe change network to 192.168.42.1/24

SvenRoederer:
So we can access the router always via 192.168.42.1 or it's final configuration. And there is no need for using 192.168.1.1 anymore.


SvenRoederer:
tested on WR1043-v1

bobster-galore:
go for it! looks promising n diminishes setup irritations.

booo:
I'm not sure if this creates more confusion. I guess in most of the fail safe tutorials the 192.168.1.1 is mentioned. But if you think this is a good idea let's go for it.

SvenRoederer:
I think, who is dealing with SSH on a command line, should be able to replace "1" by "42" of the tutorials.

booo:
You still need to know the answer - "42".

PR #498, opened 2017-12-21
activate "notunnel-uplink" in firmware-images

SvenRoederer:
This are the changes required to get PR https://github.com/freifunk-berlin/firmware-packages/pull/125 activated in the firmware.

Tested to work since some months in "SAm0815_Hedy-alpha" builds w/o problems


bobster-galore:
Since until now it was bloody hacking to use berlin.freifunk without a tunnel, this is finally a comfortable solution to use no tunnel. Pls merge.

SvenRoederer:
merged into Hedy-1.0.0-rc1

SvenRoederer:
merged in 4ff2c2d63166ef5b52dea2b26a9498320714c84d

booo:
Can we initiate a discussion about the default behaviour of the firmware? I don't think that we should disable the vpn/tunnel by default without communicating this change clearly in the community and on our websites and tutorials.

torte71:
@booo : Shouldn't we discuss this on the Berlin mailing list rather than assuming, what the users might expect?

We talked about that on our Spandau meeting and the majority had concerns about changing the default behaviour to using no tunnel: Many users know/use Freifunk-Berlin as a simple way to avoid legal issues and probably don't expect that change.
The word "default" is very tempting. I've downloaded that flavour several times by mistake, even though I am quite aware of the differences and wanted a tunnel flavour.
If we present the three flavours as "no-tunnel", "tunnel-berlin" and "vpn03" on the download page, users have to stop and consider, what this means and what they want.

What's the worst case?
If you don't want to use a tunnel, but use a tunnel-flavour, uplink won't work, unless you provide vpn keys.
If you expect a tunnel setup, but use the default (no-tunnel) image, everything looks working fine, until you get mail from a lawyers office.
I'd say, the latter will cause more trouble.

SvenRoederer:
@torte71 feel free to start a thread on the Mailinglist.

SvenRoederer:
for discussion see https://lists.berlin.freifunk.net/pipermail/berlin/2018-February/036982.html and following.

Issue #375, opened 2016-07-13
Set basic_rate for AP/ad-hoc


sarumpaet:
Do not allow clients with very bad signal to throttle down the whole AP. See
https://lists.berlin.freifunk.net/pipermail/berlin/2016-July/033682.html

SvenRoederer:
really for 0.2.0 ?
I prefer to do some real-life-test to find a proper setting.

sarumpaet:
https://github.com/freifunk-gluon/gluon/pull/467
https://github.com/freifunk-gluon/gluon/pull/810

lynxis:
I'm using those for years. I completely agree to deactivate B rates.

SvenRoederer:
please see upstream discussion 
https://github.com/lede-project/source/pull/945

SvenRoederer:
the wifi-option "legacy-rates" has been added to upstream LEDE to support this (https://github.com/lede-project/source/commit/e194e1b3c838a301178effb639804c28fe67354d)

bobster-galore:
Would the suggested solution in https://github.com/freifunk-berlin/firmware-packages/pull/106 be ok with you?

SvenRoederer:
PR https://github.com/freifunk-berlin/firmware/pull/499 seems to me like an easy and clean solution

SvenRoederer:
b-rates have been disabled by default in Hedy-1.0.0 (https://github.com/freifunk-berlin/firmware/commit/d3c50b6d6710ba09e212fc13c19df74250686b2c).
With a working migration we should do this also for Hedy-1.0.1

PR #499, opened 2017-12-25
hostapd: disable 802.11 legacy-rates by default

SvenRoederer:
In https://github.com/lede-project/source/commit/e194e1b3c838a301178effb639804c28fe67354d
the option "legacy-rates" was added to uci.wireless.radio. So on a per radio-basis the 802.11b rates can be disabled easily.
LEDE implemented this as having legacy-rates enabled by default, but we want to have them disabled by default. (see https://github.com/freifunk-berlin/firmware/issues/375)

So we change the default of LEDE. This also safes some work in changing the UCI-config on every node.
The node-operator is always free to enable these rates again by adding the legacy_rates option.

This was also tested on a 5GHz radio of a Dualband-router and did no show any side-effects in this band.

### DIFF
--- a/patches/005-hostapd_nolegacy_by_default.patch
+++ b/patches/005-hostapd_nolegacy_by_default.patch
@@ -1,0 +1,13 @@
+diff --git a/package/network/services/hostapd/files/hostapd.sh b/package/network/services/hostapd/files/hostapd.sh
+index 1f28661..378b114 100644
+--- a/package/network/services/hostapd/files/hostapd.sh
++++ b/package/network/services/hostapd/files/hostapd.sh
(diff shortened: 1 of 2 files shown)


bobster-galore:
Since it has been tested as well on dualband router without undesired effects, I think we could merge it.

SvenRoederer:
merged into Hedy-1.0.0-rc1

SvenRoederer:
merged in d3c50b6d6710ba09e212fc13c19df74250686b2c

SvenRoederer:
recently there is a PR on OpenWrt doing this upstream (https://github.com/openwrt/openwrt/pull/3654)

PR #503, opened 2017-12-29
Reduce disc-space usage on buildbots

SvenRoederer:
An approach to fix https://github.com/freifunk-berlin/firmware/issues/414 by adding an option to config.mk (SET_BUILDBOT) for handling the "IS_BUILDBOT" env of https://github.com/freifunk-berlin/buildbot/pull/42


SvenRoederer:
I hope the "SET_BUILDBOT" option is a acceptable way to enforce the same type of running the make job on buiildslaves or locally if it's desired.

bobster-galore:
I think it's a good step toward managing buildbots individually. We should merge.

SvenRoederer:
merged into Hedy-1.0.0-rc1, even there is no support from buildbots yet.

PR #526, opened 2018-03-06
wechsel auf OpenWRT-branch "master"

SvenRoederer:
Da ein OpenWRT-release 18.03 oder 18.04 geplant ist, sollten wir schon mal diesen Wechsel bei unserem Code vorbereiten.

Grundlegend funktioniert der Code, es gibt aber noch einige Bausstellen ... (https://github.com/freifunk-berlin/firmware/issues?q=is%3Aopen+is%3Aissue+milestone%3AHedy-1.1.0)


SvenRoederer:
Any concerns with this step towards the new major release?

booo:
Can you clean up the pull request a bit. Why does it contain e.g. "remove VPN03 flavor"? Maybe we can squash some commits?

Otherwise a good idea.

SvenRoederer:
I already squashed the original commits in a way, that they are atomic. When on of these upstream changes required some change on our code such commit was added and then the upstream changes were continued.
VPN03 was removed as I expect this next major release to come after the VPN03-service is shut down and this also saves some buildtime (probably 25%).

booo:
Could you please revert the change regarding the VPN03 package. The service will be in existence for at least until the end of the year and we need images for the service. Thanks a lot.

SvenRoederer:
When the next major-release will come to a releaseable state long enough before VPN03 will shut down (e.g. 3 months), I'll happily revert this commit.
but for not I'll save the buildbot-ressources, as noted before.

PR #528, opened 2018-03-09
Move tl-wr710n_v1.0 and tl-wr710n_v2.1 to non-tiny profile

torte71:
Only the tl-wr710n_v2 is 4MB device.
The tl-wr710n_v1.0 and tl-wr710n_v2.1 are not recognized as 4MB devices
and should have 8MB according to the tech specs.
https://wiki.openwrt.org/toh/hwdata/tp-link/tp-link_tl-wr710n_v2.0
https://wiki.openwrt.org/toh/hwdata/tp-link/tp-link_tl-wr710n_v1.0_us
https://wiki.openwrt.org/toh/hwdata/tp-link/tp-link_tl-wr710n_v2.1

### DIFF
--- a/configs/ar71xx-tiny.config
+++ b/configs/ar71xx-tiny.config
@@ -1,0 +1,10 @@
+CONFIG_TARGET_ar71xx_tiny=y
+CONFIG_PACKAGE_kmod-ath9k-htc=m
+CONFIG_PACKAGE_kmod-ath10k=m
+CONFIG_PACKAGE_ath10k-firmware-qca988x=m
+CONFIG_PACKAGE_ath10k-firmware-qca9888=m
+CONFIG_PACKAGE_kmod-rt2800-lib=m
+CONFIG_PACKAGE_kmod-rt2800-usb=m
+CONFIG_PACKAGE_kmod-rt2x00-lib=m
+CONFIG_PACKAGE_kmod-rt2x00-usb=m
+CONFIG_TARGET_SQUASHFS_BLOCK_SIZE=512
--- a/packages/backbone_4MB.txt
+++ b/packages/backbone_4MB.txt
@@ -54,8 +54,8 @@ olsrd-mod-nameservice
 kmod-ipip
 
 # BATMAN
-kmod-batman-adv
-alfred
-batctl
+#kmod-batman-adv
+#alfred
+#batctl
 
 # Statistics/collectd not included in 4MB image
--- a/profiles/ar71xx-generic.profiles
+++ b/profiles/ar71xx-generic.profiles
@@ -20,6 +20,8 @@ tl-wdr4300-v1
 tl-wdr4300-v1-il
 tl-wdr4310-v1
 tl-wdr4900-v2
+tl-wr710n-v1
+tl-wr710n-v2.1
 tl-wr842n-v1
 tl-wr842n-v2
 tl-wr842n-v3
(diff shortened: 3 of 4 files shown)


torte71:
- About the missing settings:

I am not sure, what you mean. After "make TARGET=ar71xx-tiny prepare" has completed, the resulting openwrt/.config contains (amongst others):
```
CONFIG_TARGET_ar71xx=y
CONFIG_TARGET_ar71xx_tiny=y
CONFIG_TARGET_ar71xx_tiny_Default=y
```
This is a new TARGET. Like ar71xx-mikrotic, the tiny images are not build together with ar71xx-generic (opposed to the earlier behaviour).

- About the qca9888 firmware:

I am not sure, if this (and the other additionally defined modules from ar71xx-tiny.config) is required for any of the devices. I'll do some test builds with those modules removed - and hope that openwrt will complain about missing dependencies, in case there are any. (In the mvebu/linksys-3200acm build, I've come across two modules, that were required in the final image generation step, but have not been automatically built as a dependency for this device).

The ath9k-htc + ath10k* + rt2* stuff sums up to 930KB (ipk size), and is in fact interesting for reducing the size.

torte71:
Build successfully completed with all additional modules removed from ar71xx-tiny.config:
```
CONFIG_TARGET_ar71xx_tiny=y
CONFIG_TARGET_SQUASHFS_BLOCK_SIZE=512
```
I'll refer to this setup as "stripped" and to the setup from my initial commit as "full".
I checked, that for all devices from ar71xx-tiny.profiles the respective images have been generated.
The only exception are tl-mr3220-v1+v2, for which the "backbone" images have not been built, but the "default" have been, so this doesn't look to me like a result of the missing modules.

The resulting *.bin files have all exactly the same size as before (with the "full" ar71xx-tiny.config), but when flashed to the device (checked on wr841-v9), I found two differences (sizes in KB, compared the outputs from "du -a /" using diff):
```
full    strip   file
13      8       /lib/modules/4.9.85/compat.ko
414     410     /usr/sbin/wpad
```
Not much, but each byte counts on those devices :). I didn't check, what size increase each single module does.

I'm OK with removing either the qca988* stuff or all the additional modules.
Providing installable modules for devices without opkg is worth a second as well.


- "backbone" is still too big
Even with the "stripped" config, the /overlay partition becomes too small - already the first step from the wizard, setting the password, doesn't work and gives "No space left on device" in logread output.

- "default" looks working (checked "stripped" and "full")
  - device gets uplink via mesh
  - gives access to inet + Freifunk to clients connected via WiFi
  - overlay partition still big enough for configuration data (checked by running wizard, restoring backup using luci+sysupgrade+manual extraction)

SvenRoederer:
regarding the options
>CONFIG_TARGET_ar71xx=y
CONFIG_TARGET_ar71xx_tiny_Default=y

when you compare with the current TARGETS these options have been included on all of them. So we should handle it here the same way.
It might be that it's more cosmetic than functional, but it keeps the files in sync

torte71:
I think we should wait for the outcome of the discussion about the future of the 4mb devices. If we decide to provide at least the sourcecode, I'll create another branch+PR with following settings:
- drop "backbone" flavor
- include "ar71xx=y" and "ar71xx_tiny_Default=y", but drop all additional modules, as they can't be installed due to the missing "opkg" anyway.

So I'll close this and see this PR only as reference.

Issue #431, opened 2017-03-09
Status: 500 Internal Server Error


everloop2:
- effects Builds: master | 0.3.0-rc1 | SAm0815/Hedy-alpha_olsr0903
- LEDE Vanilla Build tested > not effected
- first boot to FFWizard works, appears after FFWizard and later accessing: frei.funk (http://frei.funk/cgi-bin/luci) > "Status Page" ,even if no password is set
- accessing: http://frei.funk/cgi-bin/luci/admin > works and full usable
- everything else (VPN, Olsrd..) is workin > Box up and running (http://monitor.berlin.freifunk.net/host.php?h=cottbus-lausi36)

tried and did not work / already fixed: https://dev.openwrt.org/ticket/19752
- mkdir /etc/syslog-ng.d
- No need for 777, uhttpd just requires files to be world readable, so:	chmod -R o+r /www
- chmod -R o+r /www
- chmod o+x /www/cgi-bin/luci
- add package Luci-Splash

Hostname: cottbus-lausi36
Model: Arcor 803 (Easybox803, https://wiki.openwrt.org/toh/astoria/arv752dpw22)
Firmware Version: Freifunk Berlin Hedy 1.0.0-olsrd0903-alpha 9a4fd73 r3205-59508e3 / LuCI lede-17.01 branch (git-17.051.53299-a100738)
Kernel Version: 4.4.50

Syslog:
/usr/lib/lua/luci/dispatcher.lua:380: Failed to execute function dispatcher target for entry '/'.
The called action terminated with an exception:
/usr/lib/lua/luci/dispatcher.lua:380: Failed to execute function dispatcher target for entry '/freifunk'.
The called action terminated with an exception:
/usr/lib/lua/luci/dispatcher.lua:380: Failed to execute template dispatcher target for entry '/freifunk/index'.
The called action terminated with an exception:
/usr/lib/lua/luci/template.lua:55: Failed to execute template 'freifunk/index'.
A runtime error occured: /usr/lib/lua/luci/template.lua:55: Failed to execute template 'header'.
A runtime error occured: /usr/lib/lua/luci/template.lua:55: Failed to execute template 'themes/bootstrap/header'.
A runtime error occured: [string "/usr/lib/lua/luci/view/themes/bootstrap/hea..."]:150: attempt to index local 'boardinfo' (a nil value)
stack traceback:
[C]: in function 'assert'
/usr/lib/lua/luci/dispatcher.lua:380: in function 'dispatch'
[...]

PR #530, opened 2018-03-10
Set "umask 022" for "git clone" and "feed update"

torte71:
The openwrt files fetched via "git" already need the correct
permissions.

https://github.com/openwrt/luci/issues/1521
https://github.com/freifunk-berlin/firmware/issues/431

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -45,7 +45,8 @@ default: firmwares
 
 # clone openwrt
 $(OPENWRT_DIR):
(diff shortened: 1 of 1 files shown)


booo:
Please try to add a few lines to your commit message(s) that add some context. What is the problem and why does your commit solve the problem.

https://chris.beams.io/posts/git-commit/

SvenRoederer:
@booo I think it's more nice to "merge" such single commit PRs by rebasing rather than merging. This will make the history more readable by have less merge commits.

PR #541, opened 2018-04-01
add bmx and babel routing

SvenRoederer:
this will add some more routing-daemons as binary -packages ready to install. This includes BMX6, BMX7, babel and their available LuCI-interfaces.


SvenRoederer:
as discussed in https://github.com/freifunk-berlin/firmware/commit/9c00ee70ea66dabb142bc4200d8ece8a67bd40a8

SvenRoederer:
builds as expected

PR #546, opened 2018-04-05
ar71xx-tiny: reduce kernel-size

torte71:
Reduces the size of the kernel about 30KB (size of "*-vmlinux-lzma.elf").
This can lead up to 64KB more space for the rootfs partition because of the alignment of the MTD partition.
The patch is specific to the wr841 models, as the biggest part (17KB) comes from removing PHY code, that is not required for this device.
See commit messages and https://wiki.freifunk.net/Benutzer:Torte/KernelSizeDraft for details.


booo:
Please add some documentation about the new tool/feature you introduce e.g. to the README.

Maybe this should be address in the openwrt build process rather then in our makefile?

Does this result in usable images for 4MB routers?

torte71:
- Docs: OK, I'll add some documentation.
- Usable images: The backbone flavour is useable again with the additional space (batman has to be removed, as we did already in 0.3.0). OpenVPN-mbedTLS still does not fit, even with the patches from @SvenRoederer - but it's a step forward (and e.g. "tc" might be reducable as well). I didn't check "lede_wizard" with tiny-kernel and Svens OpenVPN stuff.
- Upstream changes: I doubt, that all the settings will be accepted by openwrt. "EFI_PARTITION" and "MAGIC_SYSRQ" won't be a problem, "COREDUMP" and "DEBUG_FS" might require some good arguments. But I am not optimistic, that the stripped "PHY" code will be accepted, as this reduces the amount of usable devices - I am not sure, if this kernel would run on any other devices than wr841 (which is widely used for Freifunk-Berlin). But this makes up 17KB.
As long, as only some of the changes are accepted upstream, my proposed Makefile change would still be necessary, if we want that space reduction (and it only needs some KB size increase of the kernel, that its MTD partition will suddenly grow by 32KB).

booo:
As far as I understand the problem there is no way to change the kernel configuration with the current openwrt tools/build process. The kernel options specified by the user are overwritten by the current toolchain. Why is there no option to specify kernel build options? Why does openwrt overwrite options from the user? Maybe we should fix this upstream?

torte71:
The part, that overwrites most (not all) of the kernel-options is the (openwrt) "make defconfig" step (called from our Makefile, just before my change). This step is used to switch between different TARGETs, so I can understand, why it changes that much.
But in fact, something like a (openwrt) "make changedconfig", that incorporates the additional options in a second step would be useful.
And a more relaxed or intelligent check for (openwrt's) "config out of sync" would be nice: Currently, this check sees "# FOO is not set" (even if this will finally evaluate to "n") as different from "FOO=n" and emit a warning. This was the reason for not simply "cat"ing the options to .config, but removing the settings via some globbing/sed in the Makefile.

SvenRoederer:
some of the "disabled" config-settings can be changed like I did in https://github.com/freifunk-berlin/firmware/blob/112da0a6dc9f95c41941ef8c0e991769c3855176/patches/000-ar71xx-tiny_reduce_kernelsize.patch .
I think we can change the "defconfig" for the tiny target for many (probably for all?) items this way. And  this kind of patch will have a good chance to get upstream (as @NeoRaider of @freifunk-gluon might also be interested)

bobster-galore:
Why not pulling this request?
If there is a possibility to use the old plastic, it's worth not to pollute the seas with plastic, I think.
And with https://github.com/freifunk-berlin/firmware/issues/548 there is some chance to get it work.

SvenRoederer:
@booo the official way to change the kernel-configuration is to call ```make kernel_menuconfig```

@bobster-galore  this change in the Makefile looks too hackish to me.
* this logic there is something I would never touch
* defining config-options in openwrt/.config which are not in relation to any corresponding CONFIG_*parameter, but are supposed to pop-up in the kernel-config ...

SvenRoederer:
* https://git.openwrt.org/?p=openwrt/openwrt.git;a=commit;h=2d268c07108f3d53e0295a89ee34a87b26f4e829 shows some similar settings, esp. for the PHYs.
* https://git.openwrt.org/?p=openwrt/openwrt.git;a=commit;h=80e3a16fda096ecb3e66c9def0ab6e864ea49046 suggests also an option so safe some space.

booo:
Please reopen if there is still interest in pushing this forward.

PR #562, opened 2018-05-13
config: Add tunneldigger module

pmelange:
The feeds.conf and common.config changes adds support for tunneldigger.


feeds.conf
```diff
@@ -1,6 +1,7 @@
 src-git packages https://github.com/openwrt/packages.git^56331e808f499610c7a3ee295485ad49a9d351e7
 src-git luci https://github.com/openwrt/luci.git^c1e6a3dd3ae1be749c8608ffec8d9c90362e17db
 src-git routing https://github.com/openwrt-routing/packages.git^b82d7ccd0abe042359c41229090c2ce5d4276950
+src-git wlanslovenija https://github.com/wlanslovenija/firmware-packages-opkg.git^c3ae82b7f1d411ef0cd7ce594697b4167bb00462
 src-git packages_berlin https://github.com/freifunk-berlin/firmware-packages.git^e23be3e9922896c540e8f7d25a36917ce72aff9b
```
SvenRoederer: why adding this feed?
This also defines a tunneldigger-package, we already have via the gluon-feed?

pmelange: Doesn't it make sense to take the package from the main repository and origional developers and not from a forked branch?  The two are different.  Not very different, but still somehow different.  

Also, the tunneldigger-broker which I am working with is from wlanslovenija.  They wrote all the documentation and are in my opinion more responsible for the tunneldigger project.

pmelange: After looking closer, the gluon version sources a more recent version of the client code than the  wlanslovenija version.  And the gluon init script reads more options from the uci config file.

But the gluon version does not seem to be an official fork of the wlanslovenija version.  But more like they just copied the code from them into their repository.  

I don't know any more which one makes more sense.  The official out-of-date package which we could contribute to or the non-official package which does not contribute to the official one.

SvenRoederer: > Also, the tunneldigger-broker which I am working with is from wlanslovenija. They wrote all the documentation and are in my opinion more responsible for the tunneldigger project.

https://github.com/wlanslovenija/firmware-packages-opkg HEAD (https://github.com/wlanslovenija/firmware-packages-opkg/commit/c3ae82b7f1d411ef0cd7ce594697b4167bb00462) is from 26 Jun 2017 ; https://github.com/wlanslovenija/firmware-packages-opkg/blob/c3ae82b7f1d411ef0cd7ce594697b4167bb00462/net/tunneldigger/Makefile#L6 uses tunneldigger with 297592db89c7cb302db7a23e7a4dd5ceef957354 (https://github.com/wlanslovenija/tunneldigger/commit/297592db89c7cb302db7a23e7a4dd5ceef957354) of 11 Oct 2015.
Gluon-packages is using tunneldigger https://github.com/wlanslovenija/tunneldigger/commit/d7db350011076d6a83855d412885a29a7d142b6e of 13 Dec 2017 (https://github.com/freifunk-gluon/packages/blob/6c38a5dd639ee4649bc98401c5b3bb27ff99de4d/net/tunneldigger/Makefile#L9) - just a "bit" more recent.

pmelange: I've updated the branch to include the gluon feed instead of wlanslovenija.

pmelange:
My experience with the master branch is not positive.  The changes in the pull request do not change the resulting image in any way.  Instead it only adds more packages which people can install after installing the firmware.

If needed, I would then rather keep this in a separate branch and have the buildbot make an "experimental/development" version of 1.0.1.

SvenRoederer:
> My experience with the master branch is not positive. The changes in the pull request do not change the resulting image in any way. Instead it only adds more packages which people can install after installing the firmware.

from this side, there is no difference between Hedy-1.x, master, experimental or any other branch

pmelange:
I've just given the latest master image from the buildbot a try and it seems to be working, although with a constant high load.  I think there are still quite a few issues to work out with the master branch before it could be called "not different from Hedy-1.x".  For example, the olsrd plugin version, the ESSID and BSSID in adhoc all needed manual intervention before the router worked.

The latest Sam0815 image didn't even make it though the firmware assistant.

I'm not sure how to proceed from here.  Should I change the branch to go from master (which would take potentially a long time to be released)?  Or, can I have the added kernel modules and the tunneldigger package included in the 1.0.1 release?

IMHO, adding this to 1.0.1 should not be a problem since nothing is changed in the generated firmware image.  Just new packages will be available for "experts" to install via opkg install.

pmelange:
Closed without comment?  Can you please tell me why?

SvenRoederer:
this was for sure autoclosed when I removed the orphaned Hedy-1.0.1 branch.

pmelange:
I would really like to get this into the next release.  Which branch should I base it on?  Hedy-1.0.2, next_release, master (which is unstable)?

booo:
I added tunneldigger as module to the master branch with commit e905ad570f2e64351b644cef7b21f5474c89dbae.

PR #567, opened 2018-05-27
New features for master

SvenRoederer:
* CONFIG_PKG_ASLE_PIE: improved security for binaries, seems not to make them larger
* CONFIG_PACKAGE_simple-tc: alternative, possible smaller package way of tc
* luci-app-vpnbypass: include into image by default, can help to have the users not to push all over the tunnel

I'd like to merge, any concerns?


bobster-galore:
looks nice!

SvenRoederer:
This seems to get merged a bit to quickly, as the branch "next_release" did not build ....

booo:
I would prefer one pull request per feature the next time.

The last time we added a feature to improve the security of the firmware we removed it later on. I think we should either make a real effort to improve the security of the system or don't do it at all.

Could you please be more specific/verbose about the olsr-plugin-names change? I can't find the corresponding commit in firmware-packages based on the commit or commit message.

Who requested the vpnbypass as feature? We can build it as module and users can install it on demand. I don't see why this should be part of the default package set. We have to take special care for every package that becomes part of the default package set. I don't want to add extra work for a package that is used by zero to one percent of the users. The feature set it provides is covered by the firewall app too.

SvenRoederer:
> Who requested the vpnbypass as feature? 

There was a discussion on the Mailing list some time ago, how Youtube and other streaming can be offloaded from the VPN to save some ressources.

PR #579, opened 2018-07-23
disable 802.11s mesh_forwarding by default

SvenRoederer:
By default a node must have the internal 802.11s forwarding disabled,to not interfer with the mesh-routing protocols on top.

This is done by patching the kernel-default. Also update the LuCI-frontend to stay in sync.

This was tested with OpenWrt-master, and as the patches apply to OpenWrt-18.06 too, it will work also.

### DIFF
--- a/patches/001-mac80211_mesh_no_fwding_by_default.patch
+++ b/patches/001-mac80211_mesh_no_fwding_by_default.patch
@@ -1,0 +1,18 @@
+diff --git a/package/kernel/mac80211/patches/999-mac80211-mesh-no-forwarding-by-default.patch b/package/kernel/mac80211/patches/999-mac80211-mesh-no-forwarding-by-default.patch
+new file mode 100644
+index 0000000000..5bff861737
+--- /dev/null
++++ b/package/kernel/mac80211/patches/999-mac80211-mesh-no-forwarding-by-default.patch
+@@ -0,0 +1,12 @@
++diff -ur backports-2017-11-01/net/wireless/mesh.c backports-2017-11-01.patch/net/wireless/mesh.c
++--- backports-2017-11-01/net/wireless/mesh.c	2018-07-15 15:13:56.230792586 +0200
+++++ backports-2017-11-01.patch/net/wireless/mesh.c	2018-07-22 01:36:13.299928269 +0200
++@@ -68,7 +68,7 @@
++ 	.min_discovery_timeout = MESH_MIN_DISCOVERY_TIMEOUT,
++ 	.dot11MeshHWMPRannInterval = MESH_RANN_INTERVAL,
++ 	.dot11MeshGateAnnouncementProtocol = false,
++-	.dot11MeshForwarding = true,
+++	.dot11MeshForwarding = false,
++ 	.rssi_threshold = MESH_RSSI_THRESHOLD,
++ 	.ht_opmode = IEEE80211_HT_OP_MODE_PROTECTION_NONHT_MIXED,
++ 	.dot11MeshHWMPactivePathToRootTimeout = MESH_PATH_TO_ROOT_TIMEOUT,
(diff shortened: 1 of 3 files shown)


SvenRoederer:
Any concerns with this? 
When it builds with the current master (OpenWrt 18.06.0) I like to merge this.

booo:
I don't think we should maintain two patches just to change the default of an option.

bobster-galore:
I think it's right to change the default in the driver as well as in luci, to be sure.
What would be the alternative?

SvenRoederer:
@booo indeed the patches relate to each other in some degree. But putting them into one single patch will break with our naming-conventoin. But both patches are targeted to very different upstream-projects

@bobster-galore I think there is no real alternative in setting the default to "no forwarding" as it will cause more problems to the user to have to change this setting everytime he sets up a 802.11s interface manually.

booo:
Keeping these changes, in one file or two, up to date requires work. I don't think the change is worth the extra workload.

How do we document the change properly? A user that read the documentation assumes that the option is enabled by default.  With this change and without a proper place for documentation this assumption is clearly wrong and leads to confusion.

The wizard will create the default configuration for 802.11s devices. Adding the option to disable forwarding is a minor change to the wizard code.

Why do we discuss this now? As far as I know there is no 802.11s related code in the firmware until now. Once we have code for 802.11s in the master brunch we can discuss changes to the default configuration.

SvenRoederer:
there is some code in my "staging tree" to setup 802.11s by default (https://github.com/freifunk-berlin/firmware-packages/commit/8466101dee9a5c8c186e6b05cf6360fb8031ffae) this code relies on the changes of this PR.

bobster-galore:
We had a discussion in Spandau and all 4 voted for disabling by default as we are at freifunk with olsr and not at openwrt where it should be activated by default.
Can we merge pls.

pmelange:
bobster-galore wrote in https://github.com/freifunk-berlin/firmware/commit/4155686786432adab4b56d358e4e6fbd28d0d0a9#commitcomment-30931326

> We're on freifunk, everybody can do what she likes to do! ;)
Beside that, what could be the nicest way to set a value which is for sure better suiting the freifunk environment than openwrt environment?
We could mention on wiki router page that in freifunk context mesh(11s) is always meshfwd.default = "0" cause it would bring problems with olsr on the interface.
How can we avoid to tell evryone turn it to "0" to make it work?

According to the 802.11s standard
> 802.11s requires Hybrid Wireless Mesh Protocol, or HWMP,[1] be supported as a default. However, other mesh, ad hoc (Associativity-Based Routing, Zone Routing Protocol, and location based routing) or dynamically link-state routed (OLSR, B.A.T.M.A.N.[citation needed]) may be supported or even static routing (WDS, OSPF).

https://en.wikipedia.org/wiki/IEEE_802.11s#Closely_related_standards

Changing what the standard bodies define as a default to avoid the problem that someone sets it up wrong is problematic to me.  

Now about documentation:  Our own github page has the following:
> The firmware is based on vanilla OpenWrt with some modifications (to fix broken stuff in OpenWrt itself or for example LuCI) and additional default packages/configuration settings. New features like a new network concept will be part of future releases.

https://github.com/freifunk-berlin/firmware

There is nothing broken with the standard.  We are modifying the standard which the Linux kernel and the OpenWrt is based on.  Default configuration settings I consider analogous to how the ad-hoc setup uses SSID " intern-ch13.freifunk.net" after the wizard is ran.

Where is the documentation?  Where should it go?  

Right now the OpenWrt documentation doesn't have anything about the forwarding option, but eventually it probably will.  When it does, then we will __not__ be able to point to the OpenWrt documentation as a resource for the Freifunk firmware.

SvenRoederer:
I don't see a point in "violating" the standard by just disabling the 802.11s integrated forwarding. We do not change the protocol itself, all required features (e.g. HWMP) are still there. and can be enabled by the user.

Indeed there is nothing broken with the standard setting, it just makes no sense in our freifunk-context. And that we fix with the changed default of not enabling forwarding.

Issue #534, opened 2018-03-16
Rename "default" image type


sarumpaet:
Releasing Hedy-1.0.0 with changed semantics for "default" wasn't very good, and "default" is an awful name anyways.
We should get rid of that and use a name that suits the image type better. Since we have "backbone" without the wizard already, what about using image names "wizard-notunnel", "wizard-vpn03", "wizard-tunnelberlin", "backbone" or similar?
Also see https://lists.berlin.freifunk.net/pipermail/berlin/2018-March/037260.html

niccokunzmann:
I would say "wizard-notunnel" could be "wizard-direct" or "wizard-lan" because this describes that there is not no internet but your own.
We can also remove the "wizard" word. Only a handful of geeks knows that a wizard is not always a magician.

I would suggest:
- "wan-access"
- "vpn03-tunnel"
- "berlin-tunnel"

pmelange:
I like "no-tunnel", "tunnel-berlin", and "vpn03-deprecated" along side "backbone"

SvenRoederer:
In der Mailinglisten-"Diskussion" (https://lists.berlin.freifunk.net/pipermail/berlin/2018-February/036982.html) waren jetzt keine Tendenzen für Pro oder Contra abzusehen und daher blieb es erstmal bei der bekannten Namenswahl.

Von den hier gemachten Vorschlägenwäre ich für
>  "wizard-notunnel", "wizard-vpn03", "wizard-tunnelberlin", "backbone"

SvenRoederer:
Aber das "blocker" label find ich jetzt übertrieben ...

kehugter:
> "wizard-notunnel", "wizard-vpn03", "wizard-tunnelberlin", "backbone"

+1

> "*-deprecated"

For consistency, I'd put deprecation/experimental/beta/git/etc notes outside of the name.

bobster-galore:
I think it's a nice idea to drop "default" since it doesn't mean anything.
We can have a long discussion what is more understandable to noobs either wizard or assistant but to me wizard sounds cute.

>     "wizard-notunnel", "wizard-vpn03", "wizard-tunnelberlin", "backbone"

I think wizard-* promises smthg while backbone sounds like u r left on ur own. Which is ok, I think. 

Wizard is compatible with old and new (alternative) wizard.

By dropping default we support thinking about what to choose, which is nice as well.

bobster-galore:
If 


[...]

PR #592, opened 2018-10-04
add uplink type tunneldigger to the build environment

pmelange:
This PR is dependent on https://github.com/freifunk-berlin/firmware-packages/pull/144 being accepted


SvenRoederer:
per https://github.com/freifunk-berlin/firmware/issues/534 all seem to agree in adding more image-types is a no-go.

pmelange:
When there is code to switch between uplink types and to automatically install the needed packages, then your issue with this commit will be resolved.

As nobody is doing the work, then it makes sense to have a new image name. The code is ready to be tested and used. Let's start testing

booo:
We will consolidate the image types later on in the process.

We would like to propose tunneldigger as the next standard tunnel solution for the community tunnel. Openvpn will become obsolete once we make this change. I suggest that we remove the openvpn image time once tunneldigger is ready for use.

Issue #597, opened 2018-10-16
the README.md often points to outdated pages (of OpenWrt)


SvenRoederer:
Since the remerge of LEDE and OpenWrt some of the infrastructure has changed and was replaced by new instances.
Our README.md file still links to the historic pages. These are at least:
* the Wiki-pages
* the development pages

bobster-galore:
What could be keywords to extract a list from github what to alter?

SvenRoederer:
I just read the text and found some. At least the wiki-pages need do be checked for the current counterparts.
Some improvment are in https://github.com/freifunk-berlin/firmware/pull/598, https://github.com/freifunk-berlin/firmware/pull/599

SvenRoederer:
I assume this was autocolsed somehow

PR #599, opened 2018-10-16
README: update section of valid targets and ...

SvenRoederer:
Just some update(s) of the README.

feel free to add updates to fix https://github.com/freifunk-berlin/firmware/issues/597 also.


booo:
I think there is something wrong with the spelling but who cares...

PR #607, opened 2018-10-22
README: update links to point to new openwrt wiki

booo:
The wiki we referenced is in read-only mode and will cease to exist in
the future. I update the links to point to the new openwrt
documentation.

Fixes https://github.com/freifunk-berlin/firmware/issues/597

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,7 +83,7 @@ http://github.com/freifunk-berlin/packages_berlin.
 
 ### Build Prerequisites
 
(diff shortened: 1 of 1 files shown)


SvenRoederer:
Fixing other things in the doc would have been better in https://github.com/freifunk-berlin/firmware/pull/599, but here is also fine I think

pmelange:
I have added the small changes to this branch

pmelange:
But the sub-howtos about flashing and the assistant are full of pictures:

https://wiki.freifunk.net/Berlin:Firmware:Howto:Flash
https://wiki.freifunk.net/Berlin:Firmware:Howto:Assistent

But I agree, the new howto could definately use some love.

Issue #606, opened 2018-10-22
tunneldigger config change, bind to interface


pmelange:
This applies to both ffuplink and bbbdigger.

If the node is not connect to wan, but meshes, tunneldigger will attept to build the tunnel though another the smart gw.  This creates an undesired loopback.  The solition is to add ```option bind_interface 'wan'```.

Changing the option solves this problem  But leads to another.  Even though the wan interface is not up, tunneldigger tries to connect.  This produces a lot of log entries.  

I see two pssible solutions.   [Edit, see below]

pmelange:
There is an issue at https://github.com/wlanslovenija/tunneldigger/issues/84

booo:
Please add a hotplug script that starts/stops the tunneldigger service. We did the same for openvpn.

pmelange:
OK.  So now the question is, should we keep using and patching the gluon tunneldigger (it is just a Makefile, initscript, and sample config file). Or should we just make our own tunneldigger package.

The gluon script grabs the source code from wlanslovenia.  We could do that just as easily.

Where would our version of tunneldigger go?  in firmware-packages/utils, firmware-packages/addons or firmware-packages/defaults?  Will the build env always take our tunneldigger package over the one from gluon?  Or should we name it freifunk-berlin-tunneldigger to aviod any conflicts?

pmelange:
I have a rewritten init script and a functioning hotplug script.

I'm still not sure how to package this into the firmware.  Here are my ideas:
1) Update the patch https://github.com/freifunk-berlin/firmware/blob/master/patches/101-tunneldigger_init.patch
2) Create a package similar to how freifunk-gluon does it (basically a makefile which pulls the tunneldigger code, builds the client, then adds the specialized init and hotplug script)? https://github.com/freifunk-gluon/packages/blob/master/net/tunneldigger/Makefile
3) Create a freifunk-berlin-tunneldigger-default package which contains only the init and hotplug files. (Will this properly overwrite the init script from gluon)?

All of these options do not force us to fork tunneldigger.

booo:
Can you upload your work somewhere and reference it here?

Why do you have to rewrite the init script?

[...]

PR #608, opened 2018-10-24
update tunneldigger patch to add hotplug functionality

pmelange:
This PR should be merged together with https://github.com/freifunk-berlin/firmware-packages/pull/148

An issue arrises when a router is not connected to a wan and only meshes. The wan is not connected to anything, yet tunneldigger tries to make a connection anyhow.  To aviod this, the init script has been modified to _not_ make tunnels when they are configured with the bind_interface option.  Additionally, a hotplug script has been created to create and destroy tunnels on demand as the interfaces go up and down.  The hoplug script only works with tunnel configured with the bind_interface option.

More info at https://github.com/freifunk-berlin/firmware/issues/606


pmelange:
Please don't merge this yet.  I'm working on another solution which will have it's own PR-

PR #614, opened 2018-11-03
add support for the freifunk-berlin-tunneldigger package

pmelange:


SvenRoederer:
removing the gluon-feed will stop building "simple-tc" and "l3roamd" - is that intended?

pmelange:
That was an unintended consequence.  The reason to have the feed is not commented in feeds.conf.  I can easily drop that commit and create another with comments with the reason for the feed. 

Otherwise, would you find the PR ready to merge?

pmelange:
The changes in firmware-packages have been merged since https://github.com/freifunk-berlin/firmware-packages/commit/0e335a6a09f90dbfff58dd6a354a1112cc04ccd0

An extra commit to feeds.conf finishes the update https://github.com/freifunk-berlin/firmware/commit/3ad7a0ff4aa5386328a88dbd89c92ba7b3154187

Issue #609, opened 2018-10-24
Patch 200-wifi_nolegacy_rates_by_default.patch fails with luci master


pmelange:
I just tried doing a build to test another issue which requires the latest from luci.  Unfortunately the build failed at patching 200-wifi_nolegacy_rates_by_default.patch

The feed I used is https://github.com/openwrt/luci.git^290610798a4d6ced5d0db31e1cbcad7d758e020f

```
Applying patch 200-wifi_nolegacy_rates_by_default.patch
can't find file to patch at input line 5
Perhaps you used the wrong -p or --strip option?
The text leading up to this was:
--------------------------
|diff --git a/modules/luci-mod-admin-full/luasrc/model/cbi/admin_network/wifi.lua b/modules/luci-mod-admin-full/luasrc/model/cbi/admin_network/wifi.lua
|index c0bb38030..877395a36 100644
|--- a/feeds/luci/modules/luci-mod-admin-full/luasrc/model/cbi/admin_network/wifi.lua
|+++ b/feeds/luci/modules/luci-mod-admin-full/luasrc/model/cbi/admin_network/wifi.lua
--------------------------
No file to patch.  Skipping patch.
1 out of 1 hunk ignored
Patch 200-wifi_nolegacy_rates_by_default.patch does not apply (enforce with -f)
Makefile:92: recipe for target '.stamp-patched' failed
make: *** [.stamp-patched] Error 1
```

SvenRoederer:
please check https://github.com/freifunk-berlin/firmware/commit/02a49c7991c6c502a3f5a0954daf4819b1e1f74a

pmelange:
Thanks for the pointer.  I cherry-picked the commit into my working branch, updated the feed, and am building locally.  Hopefully I'll be able to test later today.

pmelange:
This will be resolved when #615 gets merged

SvenRoederer:
btw. is this also relevant for 18.06 branch?

SvenRoederer:
can also be closed as of https://github.com/freifunk-berlin/firmware/pull/615#issuecomment-459145262

Issue #583, opened 2018-08-19
OpenWrt-base for next minor-release


SvenRoederer:
@bobster-galore once mentioned that we should consider using OpenWrt 19.x as base for the next minor-release (Hedy-1.1.x).

That idea is based on:
* OpenWrt 18.06.1 is out already, but we a far from a point we can think of a release
* building 1.0.2 on OpenWrt 18.06 will bring us again in the situation, that we have to run behind the features that that the OpenWrt-guys do for master
* what do we loose when skipping OpenWrt 18.06 and choose a well working OpenWrt-master?

SvenRoederer:
for some reason the master seems to have problems with at least the Archer C7v4, whereas the Sam0815_experimental branch (based on OpenWrt master) seems to work. see https://lists.berlin.freifunk.net/pipermail/berlin/2018-August/038213.html - let's take this as a vote for OpenWrt-master

bobster-galore:
Yes pls, I think there is no worth in using outdated software only for historical reasons.

booo:
When we started to develop this firmware one of our goals was a stable firmware. Therefore we chose a stable openwrt release as the foundation of the firmware. If we start using the master branch or some unpublished release we compromise the idea of a stable firmware.

SvenRoederer:
As mentioned before we are far away from a 1.1.x release. When we keep our current development-speed, we might have our code ready for a relase, when OpenWrt 19.x or some of its rc-versions has been published.
By skipping OpenWrt-18 and going from lede-17.x to OpenWrt-19.x  we avoid running behind the OpenWrt-release and even stay on a published stable openwrt release as the foundation

bobster-galore:
I understand the idea of stable firmware. On the other hand it seems to me that development of routers speeded up compared to early freifunk times, when the same router model ran for a longer time.

I think it's desirable being also able to support recent routers (e.g.archer c7 v4, cpe210v2, wr1043v5). What about developing on OpenWrt-19.x and if we find we are ready to release we check if it's already stable or backport?
@torte71 @pmelange how u mean?

booo:
I'm ok with using 19.x for one of the next releases.

What are the open issues for a next release? Why a we so far away from a next release? Is there a ticket for a next release?

bobster-galore:
Are the next steps?
1.0.2 sw-update Lede, DNS-Server out https://github.com/freifunk-berlin/firmware/issues/551 
[...]

PR #615, opened 2018-11-03
Update to the Luci feed master branch

pmelange:
Included are changes to the patches we use.


pmelange:
seeissue #609

SvenRoederer:
switching to Luci-master should in turn cause switching OpenWrt and the other feeds to master. This will lead to https://github.com/freifunk-berlin/firmware/issues/583

pmelange:
Yes, this should be merged in the future (maybe January).  When there is a 19.x fork, this PR could come in handy.

Maybe we should keep an upstream_master branch for proactive changes to the upstream code

SvenRoederer:
> Maybe we should keep an upstream_master branch for proactive changes to the upstream code

I made some scripts, which automatically updating branch "daily/upstream-master" to the recent changes of the master-branches we rely on.
So we can collect such "cutting edge changes" there.

As of this, I close this one here

PR #634, opened 2018-12-06
profiles: add target mir3g (Xiaomi Mi WiFi R3G)

pmelange:


pmelange:
I built the image locally, and it works.  There are still some upstream issues with the 2.4Ghz driver, but otherwise this target looks promising.

SvenRoederer:
What are the issue on 2G4Hz you mention?

pmelange:
I dug a little bit deeper into the issue.  It seems like there are connectivity problems when the USB3.0 kernel modules is loaded and strange disconnects at 2.4Ghz.  This problem is present in the original firmware, Padavan, and OpenWRT.

Here are the forum threads from the OpenWRT ToH page:
https://forum.openwrt.org/t/xiaomi-wifi-router-3g/5377
https://forum.openwrt.org/t/xiaomi-wifi-router-3g-18-06-x-wifi-issues-2-4ghz-5ghz/20169
https://forum.openwrt.org/t/xiaomi-wifi-router-3g-18-06-x-feedback-and-help/19840/3

My test router has been up for 3 days now, meshing with 802.11s (only 2.4Ghz neighbours) to see if there are any problems.  So far it's working fine.

SvenRoederer:
as long as the 5GHz seems so unstable (found a note about loss of traffic every 3 minutes) we should clearly note it a "unstable support" / "non-production"

pmelange:
@SvenRoederer Would you like to test with one of these routers?  Kaya told me yesterday that the router is working wonderfully.

SvenRoederer:
no thanks.
is the wifi-part also included in wonderfully?

pmelange:
Wifi works with the generated image, if that's what you mean.

SvenRoederer:
sounds good. This board should be tested with OpenWrt-master base ...

pmelange:
I have it running with the alpha-test-tunneldigger images (built with ImageBuilder).  It's working fine so far.  This router is not very expensive and has a good amount of power.  It might be possible that this router could become the next recommended freifunk-berlin router.

PR #644, opened 2018-12-17
Create ipq40xx-generic.config

RobinBrandt87:
add config for gl B1300


SvenRoederer:
@RobinBrandt87 @pmelange Why have PR #644 and PR #645 not been combined into a single one PR? This would help to keeps thinks logically together.
These recent merges look like "merge at any costs", This might be ok for branches that will be cleaned later, but looks bad on master-branch

configs/ipq40xx-generic.config
```diff
@@ -0,0 +1,16 @@
+CONFIG_TARGET_ipq40xx=y
+CONFIG_TARGET_ipq40xx_generic=y
+CONFIG_TARGET_ipq40xx_DEVICE_glinet_gl-b1300=y
+CONFIG_TARGET_BOARD="ipq40xx"
+CONFIG_TARGET_ROOTFS_INITRAMFS=y
+CONFIG_TARGET_INITRAMFS_COMPRESSION_NONE=y
+CONFIG_SDK=y
```
SvenRoederer: Why has this been defined here? It's already in common.config

configs/ipq40xx-generic.config
```diff
@@ -0,0 +1,16 @@
...
+COBFIG_PACKAGE_ath10k-firmware-qca4019-ct
+CONFIG_PACKAGE_kmod-ath10k-ct=m
+CONFIG_PACKAGE_kmod-usb-dwc3=m
+CONFIG_PACKAGE_kmod-usb3=y
+CONFIG_PACKAGE_kmod-usb-dwc3-of-simple=y
+CONFIG_TARGET_ARCH_PACKAGES="arm_cortex-a7_neon-vfpv4"
+CONFIG_PACKAGE_wireless-regdb=m
```
SvenRoederer: is this really needed?  This package should be pulled be the wifi-drivers as dependency

RobinBrandt87: i dosn't found his packages in the common so i add it to the ipq40xx specific because there are also in the openwrt config for ipq40xx Target. The only packages there are actually not needed are the fritz packages.

pmelange:
@SvenRoederer what would you suggest to do now.  Revert the changes?  Force push the commits away?

SvenRoederer:
when you are quick, that's an option.
then you can also integrate the removal of setting CONFIG_SDK

pmelange:
I have reverted #644 and #645. I tried doing a force push, but that is not allowed on the master branch (or I don't have the permission).

I am working on a branch called staging/upstream_master for people who want to play with daily snapshots.  We can always port changes to master when they are ready.  For example, adding the new LuCI package "luci-app.opkg" to common.config.

pmelange:
@RobinBrandt87 I suggest making a new branch from master and to pull in the changes you need into one branch.  

I'm sorry that I jumped the gun on this one.

PR #645, opened 2018-12-17
Create ipq40xx-generic.profiles

RobinBrandt87:
add gl-B-1300


pmelange:
you might want to submit a patch to https://github.com/freifunk-berlin/buildbot/blob/master/masters/master/master.cfg

PR #652, opened 2019-01-02
Adapt to removed VERSION_NICK

SvenRoederer:
redefine Version-strings as the VERSION_NICK-option was removed upstream

* DIST = Freifunk Berlin
* NUMBER = common name-tag (Kathleen, Hedy, Develop, ...)
* CODE = will be set by Makefile from git-version with release tag (v0.3.0, v1.0.0, ...) or commit-id (as fallback)

This keeps the data-fields separate and give them a very specific meaning.

addresses: #515, #411


SvenRoederer:
This idea is around since Mid November 2018. If nobody vetoes I'll merge this by the end of the month.

pmelange:
I don't have an opinion about this in either way.  This could be merged even earlier than by the end of the month.  Maybe even next week.

PR #655, opened 2019-01-13
README: link to the wiki-page for enduser relevant informations

SvenRoederer:
As the Wiki is more up to date in this regard, just put a link in the Readme.
This also shortes the Readme a bit and helps to keep a focus on the relevant facts of the buils-system / firmware

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,15 +30,7 @@ New features like a new network concept will be part of future releases.
   * Linux 4.4.167
   * OLSR 0.9.0.3 (downgraded for BBB-VPN compatibility)
   * B.A.T.M.A.N. 2016.5 (with patches of 2019.0)
-* custom package lists for different settings
-  * "default" variant includes ffwizard, openvpn, BATMAN
-  * "default_4MB" like the "default" variant, but excludes public router statistics page (luci-mod-freifunk), monitoring (collectd), BATMAN to fit into boards with 4MB flash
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,71 +1,19 @@
...
-  * 80.67.169.40 (www.fdn.fr/actions/dns)
-  * 194.150.168.168 (dns.as250.net)
-  * 2001:4ce8::53 (as250)
-  * 2001:910:800::12 (french data network - http://www.fdn.fr/)
 
 ## Contact / More information
 
-The firmware wiki page is at: https://wiki.freifunk.net/Berlin:Firmware
+More user relevant information about the firmware are on the wiki page at: https://wiki.freifunk.net/Berlin:Firmware. There you can also find the 
+* [ReleaseNotes](https://wiki.freifunk.net/Berlin:Firmware/v1.0.1)
```
pmelange: It would be good to have a releases wiki page which links to all the releases itself.  Then we won't have to update this link anymore.

SvenRoederer: There is a Releases-chapter (https://wiki.freifunk.net/Berlin:Firmware#Releases) to which we can link. this section can in turn link to the individual releases, but I was not willing to edit the official firmware-page already.

But I feel you also like the idea, of splitting this of.

SvenRoederer: An overall releasepage is a good idea.
But having to update some minor lines in the documentation seems good in preparation of a release, this might help to review the whole document from release to release.

SvenRoederer:
So any vetoes for merging this?

PR #663, opened 2019-02-07
improve building on Buildbot and friends

SvenRoederer:
Adds a new config-file "common-autobuild.config" which holds OpenWrt config-options which should be included when building with env "IS_BUILDBOT=yes". So we get a clean solution to add options which might only be used during automatic builds.

This new file now holds the 
* VERSION_REPO which was never useful on local builds
* AUTO_REMOVE, BUILDBOT which reduces the disk-space needed on during build. So we don't need to remove some files manually during the build


SvenRoederer:
currently the build fails for openwrt18.06 based branches as of https://github.com/openwrt/packages/pull/7807.

pmelange:
The feeds.conf file points to a forked repo.  I don't think we should merge this to master yet.  Is there some other way?

SvenRoederer:
this changed feed deal with a build-problem (see https://github.com/freifunk-berlin/firmware/pull/663/commits/0589b88459457e5cbb4885acb310339842ef0179).
We can:
* wait with merging this
* add the mentioned patches to our buildprocess, till upstream is fixed

pmelange:
Can you provide links to the commits in the SvenRoederer/openwrt-packages which we would need to implement?  I prefer option "the mentioned patches to our buildprocess, till upstream is fixed"

SvenRoederer:
Just seen that setting CONFIG_BUILDBOT=y will cause building much more packages of the OpenWrt-core, probably all packages not explicitly deselected. This seems to be the reason why we see some buildfailures of libreswan (http://buildbot.berlin.freifunk.net/builders/ramips-mt7621/builds/1191/steps/shell/logs/stdio); fixed with https://github.com/freifunk-berlin/firmware/commit/694f2b7a5c813c8eba044a8d9701ea517b35e805.

Should we drop this feature at all or just do not set BUILDBOT=y?

SvenRoederer:
alternatively, I assume that we can explicitly disable the CONFIG-options that we do not want. But this seems to be more maintenance than it's worth.

pmelange:
> currently the build fails for openwrt18.06 based branches as of [openwrt/packages#7807](https://github.com/openwrt/packages/pull/7807).

Has this issue now been solved, or to put it differently, is this ready for 18.x based builds?

SvenRoederer:
the build-failure caused by jool is gone, even there might still be something "hiding" inside this package.
But all recent builds of the branch passed and I explicitly tested the "foobar"-slave with ar71xx-target.

PR #669, opened 2019-03-14
common.config: add packages needed to support UMTS sticks (modules only)

pmelange:
An active member of the Freifunk community asked that we build the needed packages to support UMTS sticks.  After a test build, the router successfully connects to the internet and runs successfully.  It would be great to be able to include this into the master branch.

None of the packages are included directly in the images, so there is no image bloat.

Here are some Howto's to set up UMTS
* https://josefsson.org/openwrt/dongle.html
* https://openwrt.org/docs/guide-user/network/wan/wwan/3gdongle
* http://www.panticz.de/node/670

There will be a wiki page coming once all configuration issues have been resolved.


SvenRoederer:
Would it make sense to create an uplink-package for configuring this via the wizard?

pmelange:
Using UMTS is analogous to wan, not ffuplink.  It would be similar to the changes needed when a user wants to use wifi-wan.

SvenRoederer:
Right, I was just thinking of something like the Uplink-packages for different WAN-setups. This might be used for DHCP, static Address, PPPoE, 4G/5G.

pmelange:
The freifunk user would need something like this:

* the ability to save various wan connectivity types (wired ethernet, wireless, umts, and / or pppoe)
* a method to be able to select which wan type to use
* the ability to turn on and off bbbdigger

Let's make a new issue to request and discuss this.

matthias-klose:
Greatly appreciated.  We should recognize the growing demand for mobile lte based freifunk uplinks in combination with unlimited day based lte flats by event organizers and such.

pmelange:
Like I mentioned above, we should create a new issue for the support of multiple wan types.  

Some questions that would need to be answered are
* What kind of wan types do we want to support (dhcp, pppoe, umts, wifiwan, etc...)?
* How can we make a framework so that new wan types can be developed and added to the already existing ones?
* How does one configure the wan type?  Command line options?  config file options?  Luci page?
* How does one change "on the fly" the current wan type?
* Does it make sense to support multiple simultaneous wan types and to spread the traffic across all of them (check out openWRT's mwan3 package)

PR #672, opened 2019-03-30
add a Template for bugreports

SvenRoederer:
such a template might help the reporter to provide as much helpful information as possible.

you can see how it looks like in my personal fork of the repo.


SvenRoederer:
To avoid bug-reports like (https://github.com/freifunk-berlin/firmware/issues/683) I'll merge this

PR #693, opened 2019-06-04
Minify ipq44xx-config

SvenRoederer:
This PR started as fix for the IPQ40xx target (122728d), but finally removes some config-options which seems to be not needed to build the current list of boards.

I kept the commits atomic, that they can be reverted individually when needed.


SvenRoederer:
did some tweaks and ready to merge again (https://buildbot.berlin.freifunk.net/builders/ipq40xx-generic/builds/70)

pmelange:
Does this also cover the changes covered in #697 and #699? If not, is there anything in those pr's which would make sense to include in this branch?

pmelange:
Actually just #697 as the other pr is not directly related

RobinBrandt87:
as you can see in https://github.com/openwrt/openwrt/blob/master/target/linux/ipq40xx/image/Makefile 
the Packages fritz-tffs and fritz-caldata needed. The Package u-boot-fritz4040 is not longer needed since 
https://github.com/openwrt/openwrt/commit/634c733065256979835ca3d7218ae8203a693683

RobinBrandt87:
with https://github.com/freifunk-berlin/firmware/pull/699 we can build Images for the Fritzbox 4040

SvenRoederer:
@RobinBrandt87 
* as we can see in the log of build 67 (http://buildbot.berlin.freifunk.net/builders/ipq40xx-generic/builds/67), not enabling u-boot-fritz4040 will cause the build to fail.
* if the fritz*-packages are needed for the FB4040, please include them in PR #699. This will keep the changes in a logical group and will make later review much more easy

@pmelange 
as the build of the all enabled boards (yeah, it's only one currently) we should merge this one and make an other commit to add the FB4040 support.

PR #697, opened 2019-06-13
Update ipq40xx-generic.config

RobinBrandt87:
repair config  for Glinet B1300 and Fritzbox4040


SvenRoederer:
This seems to be a duplicate or very similar to PR https://github.com/freifunk-berlin/firmware/pull/693.

RobinBrandt87:
tested on B1300 and F4040 it works so we can start build images

pmelange:
@RobinBrandt87 are you OK with us using the changes in #693 instead? Do they seem correct to you and can you test the image made by #693 (build 70)?

PR #698, opened 2019-06-13
Create 006-Fritz4040.patch

RobinBrandt87:
Disabel eva.bin for Fritzbox 4040 
todo: fix eva.bin


SvenRoederer:
For what reason we need to disable the eva.bin?

PR #700, opened 2019-06-13
Update series

RobinBrandt87:
add Fritzbox 4040 patch


SvenRoederer:
can you please add the commit to PR https://github.com/freifunk-berlin/firmware/pull/698 "Create 006-Fritz4040.patch"? They both belong logically together and should be kept in one PR together.

Issue #690, opened 2019-06-01
upgrading a remote node will not make it connecting to the BBB again (tunneldigger)


SvenRoederer:
### Bug report

**An upgraded node is missing the tunneldigger package**

When upgrading a remote node via BBB-VPN was working painless with pre-Hedy releases. As the required OpenVPN packages have been included in all images a remote updated node was reconnecting to BBB-VPN by itself.
Since we switched to modular uplinks and an optional alternative BBB-VPN via tunneldigger the old situation can only be archived when using a matching "tunneled uplink" package. When using a no-tunneled image the BBB-VPN link can only be restored after a site-visit an installing tunneldigger / OpenVPN manually.

It would be great to have a working BBB-VPN connection after an update.

pmelange:
This issue has been discussed at the weekly cbase meeting and various proposals have been discussed.  Additionally, if the node uses the openvpn variety (luckily there are only about 12 users of this old service left) to connect to the bbb-vpn and notunnel or tunneldigger image is used, the same problem arises.

SvenRoederer:
In fact that's something which that closely relates to the flavors of images prebuilt. Probably the attendedsysupgrades, mentioned in https://github.com/freifunk-berlin/firmware/issues/22#issuecomment-498350409, are a smart solution.

pmelange:
This has been implemented in https://github.com/freifunk-berlin/firmware/pull/701 but is also dependant on https://github.com/freifunk-berlin/firmware-packages/pull/189

PR #701, opened 2019-06-14
Add tunneldigger to the default and backbone images

pmelange:
This addresses issue #690 where a node which has been set up to
use the bbbdigger package cannot connect to the bbb-vpn server after
a firmware upgrade.

Additionally, this has been agreed upon at the firmware planning meeting 12.06.2019 https://pad.freifunk.net/p/berlin-2019

### DIFF
--- a/packages/backbone.txt
+++ b/packages/backbone.txt
@@ -73,3 +73,6 @@ collectd-mod-rrdtool
 collectd-mod-ping
 collectd-mod-uptime
 collectd-mod-memory
+
+# Tunneldigger
+freifunk-berlin-tunneldigger
--- a/packages/default.txt
+++ b/packages/default.txt
@@ -83,3 +83,6 @@ collectd-mod-rrdtool
 collectd-mod-ping
 collectd-mod-uptime
 collectd-mod-memory
+
+# Tunneldigger
+freifunk-berlin-tunneldigger


pmelange:
This commit has been cherry-picked into Gonzo-1.1.0-rc1. It would be great if these changes were merged into the master branch.

pmelange:
This PR is dependant on https://github.com/freifunk-berlin/firmware-packages/pull/189 being merged as well.

SvenRoederer:
I personally don't like this idea.
Beside this it looks like this change will enable the bbbdigger link on all these nodes (https://github.com/freifunk-berlin/firmware-packages/blob/master/addons/freifunk-berlin-bbbdigger/files/postinst.sh#L46). I don't think this is intended

pmelange:
But the package freifunk-berlin-bbbdigger is not installed on any of the images.

SvenRoederer:
Ah, got it ...
But does adding it to the "backbone"-image really makes sense? As backbone implies it's a node with real backbone-connectivity, so a backbone-tunnel might not be useful there.

pmelange:
I feel that the name 'backbone' is an inadequate name for the image. It should be something more like 'expert mode' or 'manual configuration'. I personally have used it for a core router where using the wizard didn't make sense. That same router also connects to the bbbvpn.

pmelange:
Also, there is at least one node on the backbone which also has a link to the bbb-vpn (mod77uplink.olsr).  This could be useful for situations where the link to the real backbone goes down and then the only way to access the node would be through the bbbvpn.

SvenRoederer:
Using a unique single example of mod77uplink seems not a strong argument.

regarding the naming: I think there is an issue dealing with naming changes. I think now it the right time to change the naming.

SvenRoederer:
Please check https://github.com/freifunk-berlin/firmware/pull/713 regarding the naming-change. I had a change ready in my branch already.

packages/backbone.txt
```diff
@@ -73,3 +73,6 @@ collectd-mod-rrdtool
 collectd-mod-ping
 collectd-mod-uptime
 collectd-mod-memory
+
+# Tunneldigger
```
SvenRoederer: can you note like "# Tunneldigger to resume VPN after flashing"

packages/default.txt
```diff
@@ -83,3 +83,6 @@ collectd-mod-rrdtool
 collectd-mod-ping
 collectd-mod-uptime
 collectd-mod-memory
+
+# Tunneldigger
```
SvenRoederer: see my comment for the backbone-image

Issue #602, opened 2018-10-18
buildslaves running out of disc-space (slave foobar)


SvenRoederer:
With recent OpenWrt versions from master the disk-space for a build seems to have grown again. Buildslave foobar regularly fails for target ar71xx-generic (https://buildbot.berlin.freifunk.net/builders/ar71xx-generic/builds/941).

I see the following options:
* disable this slave completely
* exclude this slave from ar71xx-generic target
* building less packages
* only build for the boards we officially support (CONFIG_TARGET_MULTI_PROFILE)
* restructure the makefile / buildprocess 
* use "CONFIG_AUTOREMOVE" of OpenWrt which cleans up the build-directory after a package is build

A similar issue we had already: https://github.com/freifunk-berlin/firmware/issues/414

SvenRoederer:
I tested with "CONFIG_AUTOREMOVE" enabled an the problematic slave passes the build (https://buildbot.berlin.freifunk.net/builders/ar71xx-generic/builds/945).

As we get "IS_BUILDBOT" set on automatic builds, it might be an option to extend the Makefile by some code including a "configs/common-buildbot.txt" with options added to the process

SvenRoederer:
A way to set "AUTO_REMOVE" is shown in PR https://github.com/freifunk-berlin/firmware/pull/663

SvenRoederer:
Issues like this should be considered when working on #804 .

PR #702, opened 2019-06-14
configs/common.config; build all kmods by default

pmelange:
As discussed at the firmware planning meeting on 12.06.2019 there was
a consensus to build all the kernel modules instead of just a select few.

see https://pad.freifunk.net/p/berlin-2019

All the kmods together take up 14MB space.


SvenRoederer:
I don't think, that we need to build kmod-video* or kmod-capi* stuff, but the maintenance-overhead in managing individual CONFIG-options per module is not worth the saved bytes.

SvenRoederer:
But with this change, we even more run into the problem if issue https://github.com/freifunk-berlin/firmware/issues/602 caused by to large buildspace needed for some buildslaves.

pmelange:
But if we get to a unified image, then we will start saving hundreds of megabytes. :)

One option would be to build all the kmods only for releases.  This option was shortly discussed.  I'm not much of an Makefile expert.  What do you think the possibilities are of doing that?

SvenRoederer:
For sure we can rewrite the Makefile to use other sets of configs for relases. Bur I'm strongly against this, as it will bring a big danger of having problems with releases, that have not been seen in unstable builds as of some sideeffects.

pmelange:
Then lets concentrate on the unified image, save hundreds of megs, and let all the kmods always be built

SvenRoederer:
In buildlog http://buildbot.berlin.freifunk.net/builders/ar71xx-generic/builds/1433 you can check on the mentioned problem of the buildspace.

SvenRoederer:
In buildlog http://buildbot.berlin.freifunk.net/builders/ar71xx-generic/builds/1433 you can check on the mentioned problem of the buildspace.

PR #713, opened 2019-06-19
rename image-flavors

SvenRoederer:
In https://github.com/freifunk-berlin/firmware/issues/534 there was some discussion about meaningful names.
Even there is no progress on having a unified image, we can rename "default" and "backbone" already.


SvenRoederer:
Can someone take care of the "backbone" name?

Please rebase this PR, as both commits would be so far apart from each other on the time-perspective.

SvenRoederer:
This was merged in https://github.com/freifunk-berlin/firmware/commit/72667f2f65164c542e8cddf2feca04bfbba958b7 and later approved to stay in master (https://github.com/freifunk-berlin/firmware/issues/534#issuecomment-617367339)

PR #717, opened 2019-07-18
Minor build changes

SvenRoederer:
This improves some small pieces of our build-system:

* gitignore: only match for /openwrt/
* Makefile: disable built-in suffix-transformation rule
* remove patches/.gitignore

Please do a rebase merge, to have this as 3 separate and independent commits

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,6 @@
 .stamp*
 patches/*~
 bin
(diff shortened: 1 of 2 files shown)


SvenRoederer:
As it passed the build, I'd like to squash it soon.

PR #733, opened 2019-10-20
packages: add mesh-flavor (OpenWrt Master)

SvenRoederer:
This flavor provides a mesh-only node with full capabilities for 4MB devices.

This is the feature-list:
- console only
- includes B.A.T.M.A.N.
- no Uplink / VPN (no firewall)
- OLSRd with watchdog

### DIFF
--- a/packages/mesh_4MB.txt
+++ b/packages/mesh_4MB.txt
@@ -1,0 +1,42 @@
+-kmod-usb-core
+-kmod-usb2
+-kmod-usb-ledtrig-usbport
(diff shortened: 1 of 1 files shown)


SvenRoederer:
After Merging https://github.com/openwrt/luci/pull/2817, this package-list will result in a comfortable amount of free squashfs-space

SvenRoederer:
merged into master (openwrt-19.07 based) with 2b86bce88b84138d948dd87e80f4ecb41d9eedf2

PR #739, opened 2019-12-04
Build all profiles to make building any board with imagebuilder work

SvenRoederer:
Enable config-option "CONFIG_TARGET_MULTI_PROFILE" & "CONFIG_TARGET_ALL_PROFILES" to make OpenWrt-buildsystem select all packages that any board is depending on.
Up to now we had to manually enable all packages (esp. kernel-packages) that a specific board depends on. For the official supported boards (listed in the TARGETS profile-file) we find such dependencies during compile-time. For "unofficial" boards such packages will no be build automatically. So the user interested in using an unofficial board has to compile the whole system himself.
As we set "CONFIG_ALL_KMODS" in #702 this problem seems not relevant for recent builds anymore, as long as this config-option is set. For the tiny-targets we will likely remove this setting in favor of saving some bytes and this will raise up the initial problem again.

This addresses #738 .


SvenRoederer:
Only the commit ```configs: build with MULTI_PROFILE & ALL_PROFILES``` is relevant. the 2 other commit are just to check the build. They remove the "KMODS_ALL" and the patch "LuCI PR2817" which breaks the build.

SvenRoederer:
Compile-tested: all buildbot-targets, mvebu-cortexa9

SvenRoederer:
merged in 04c3b4f96d7a0d1ceeeaa79f08c3b9e061bbe5b7

Issue #736, opened 2019-11-25
Request Firmware support for Netgear N600 (WNDR3700v2)  


xa-y-ax:
can we support the Netgear N600 (WNDR3700v2)?
https://openwrt.org/toh/netgear/wndr3700

SvenRoederer:
As I can see this router-family is supported :
* ath79: add WNDR3700 and WNDR3700v2 (https://git.openwrt.org/?p=openwrt/openwrt.git;a=commit;h=4dc7cb792a8e34b6ad97c7b8058c70311a06bd38)
* ramips: add factory image for WNDR3700v5 (https://git.openwrt.org/?p=openwrt/openwrt.git;a=commit;h=2fdd02cc7c3516c2cce2ace2c7b539e0821ecbe8)

This support seems also avail on OpenWrt-19.07 branch, so let's try ...

SvenRoederer:
btw. you can use the current imagebuilder to generate your own firmwareimages without running the compiler, see https://github.com/freifunk-berlin/firmware#customizing-make

xa-y-ax:
Thx! :)

pmelange:
There is a pull request to add this to the Hedy 1.0.X branch.  https://github.com/freifunk-berlin/firmware/pull/737 and is currently being built on buildbot http://buildbot.berlin.freifunk.net/#/builders/4/builds/18

After the build is finished and you test that it works, can you please leave a comment on PR #737? 

When the master branch builds again, another PR can be easily put together.

xa-y-ax:
Unfortunately the imagebuilder throws some errors - shoud i close this issue and switch to PR #737 ?

SvenRoederer:
do you have the error-message on hand?

xa-y-ax:
yep - but you already replied in PR #737 ...

Collected errors:
    opkg_install_cmd: Cannot install package kmod-leds-wndr3700-usb.
    Makefile:150: recipe for target 'package_install' failed
    make[3]: *** [package_install] Error 255
    Makefile:110: recipe for target '_call_image' failed
    make[2]: *** [_call_image] Error 2
    Makefile:206: recipe for target 'image' failed
    make[1]: *** [image] Error 2
    make[1]: Leaving directory
    '/home/xayax/firmware/imgNT0enz/freifunk-berlin-imagebuilder-1.0.4-ar71xx-generic.Linux-x86_64'
    Makefile:204: recipe for target '.stamp-images' failed
[...]

PR #741, opened 2019-12-07
Add Support for Netgear WNDR3700

SvenRoederer:
This will add support for the Netgear WDNR3700 v1,v2 and v5 as requested in Issue #736 .

This PR also includes a patch backporting the change of the V5-boardname from OpenWrt-master.


SvenRoederer:
I would like to locally add https://github.com/openwrt/openwrt/commit/3b99d6763996d4cf8d59361f25046cc8086da332 to the PR. Including this renaming will save us some work / migration when we go to the next OpenWrt branch. As this board is not officially supported yet, there is no history we need to care about.

SvenRoederer:
As nobody seems capable to test this on real hardware, I'll close this soon unmerged.

PR #743, opened 2019-12-15
Add support for GLiNet AR750 (ath79)

SvenRoederer:
This backports the support from OpenWrt master. This board is officially supported in OpenWrt-19.07 for ar71xx only, but as OpenWrt will switch to ath79 on next release, it will prevent some potential hassle when we add this board for ath79 initially.

Compile and run-test passed.

This patch has been sent upstream, but not accepted yet (http://lists.infradead.org/pipermail/openwrt-devel/2019-December/020645.html).

### DIFF
--- a/patches/openwrt/0004-ath79-add-support-for-gl-ar750.patch
+++ b/patches/openwrt/0004-ath79-add-support-for-gl-ar750.patch
@@ -1,0 +1,256 @@
+From 76a20dc00fb6889acac4bcd7daa58643014cb3f0 Mon Sep 17 00:00:00 2001
+From: Luochongjun <luochongjun@gl-inet.com>
+Date: Fri, 23 Aug 2019 17:05:57 +0800
+Subject: [PATCH 1/2] ath79: add support for gl-ar750
+
+This patch supports gl-ar750, which was previously supported by ar71xx.
+
+Specification:
+- SOC: QCA9531 (650MHz)
+- Flash: 16 MiB (W25Q128FVSG)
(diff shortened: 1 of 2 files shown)


SvenRoederer:
@Akira25 did you mention this request https://lists.berlin.freifunk.net/pipermail/berlin/2020-March/040199.html ?

SvenRoederer:
> This patch has been sent upstream, but not accepted yet (http://lists.infradead.org/pipermail/openwrt-devel/2019-December/020645.html).

It turned out, that OpenWrt will not backport this for strategical reasons (http://lists.infradead.org/pipermail/openwrt-devel/2019-December/020857.html)

Akira25:
@SvenRoederer Yep. That is was.

The device should be available natively in the next-branch by now, right?

SvenRoederer:
The next-branch (based on OpenWrt-master) has upstream support for this board and will make it avail in the future..
This is a backport for OpenWrt-19.07 to support this board in our next major release.

SvenRoederer:
@Akira25 as you mentioned, you have access to the hardware for independent test, I would be happy, is you can do it.

This PR is included in builds of branch "SAm0815-experimental_master" which you might find on firmware.berlin.freifunk.net

SvenRoederer:
after switching to OpenWrt-20.xx (73f088a3bd76fee9c6b6d8380c7132792adcb82a) the backport of the board-support is not needed and adding the profile was cherry-picked in 367dcc5

Issue #696, opened 2019-06-12
Openwrt 19x: ath10k-ct wifi-driver does not support 802.11s in wave1 devices


SvenRoederer:
### Bug report

**The ath10k-ct seems not to support 802.11s mesh and ap-station in parallel**

In https://github.com/openwrt/openwrt/commit/61b5b4971e7d6aab6b0695997c3b5aaf73c53b5f the default driver for ath10k was switched to the "Caldera Tech" version. This driver currently fails to setup a 2 VIF-configuration of 802.11s and AP, which is a common default for our routers.

**Firmware Version:** all based on recent OpenWrt master

SvenRoederer:
This is a follow up from: https://github.com/freifunk-berlin/firmware/issues/583
 
1. from: https://github.com/freifunk-berlin/firmware/issues/583#issuecomment-500386955
```iw list``` on an ath10k device still gives the following results (5Ghz):
```
	valid interface combinations:
		 * #{ managed, P2P-client } <= 16, #{ P2P-GO } <= 3, #{ AP } <= 16, #{ IBSS } <= 1,
		   total <= 16, #channels <= 1, STA/AP BI must match, radar detect widths: { 20 MHz (no HT), 20 MHz, 40 MHz, 80 MHz, 80+80 MHz, 160 MHz }
```
2. similar issues are reported by Gluon
* https://github.com/freifunk-gluon/gluon/issues/1584
* https://github.com/greearb/ath10k-ct/issues/75

SvenRoederer:
Some time ago I did a test (on a Turris Omnia) and replaced the ath10k drivers by their non-ct counterpart. As far as I remember this gave a working Mesh-Ap setup without any other change required.

So it seems to be a thing of how to change the build to include the non-ct drivers.

mweinelt:
Ath10k-ct only supports 802.11s on Wave2 chips, what hardware is this bug report related to?

pmelange:
If we want to switch back to the non-ct drivers, it looks like we will have to patch at least the following files:

firmware/openwrt/target/linux/ar71xx/image/*.mk
firmware/openwrt/target/linux/ath79/image/*mk
firmware/openwrt/target/linux/ipq40xx/Makefile
firmware/openwrt/target/linux/ipq806x/Makefile
firmware/openwrt/target/linux/lantiq/image/Makefile
firmware/openwrt/target/linux/mvebu/image/cortex-a9.mk

[...]

PR #765, opened 2020-03-05
implement Gluon build-framework

SvenRoederer:
This PR is a basic implementation of the gluon build-framework for our build-process. This idea was mainly driven by looking for a way to solve Issue #696, as our current process provides no option to define board-specific packages. See ```targets/ath79-generic``` for an example to select ath10k drivers in place of upstream-selected ath10k-ct.

The framework also brings a feature to check that all required CONFIG_options are set and packages are enabled. This way I found some faults in OpenWrt-config we used (4c975895c1c802a9f9b3a64ae4db22f4b7a15af8, a8bae2034c09fa5b568ee5afc6b108a1e949bb1b, 572abd8a8503d41a6527e9387025f2d57b35e92b).

The code is yet not a full replacement of the existing build-scripts, but it demonstrates the principle and generates usable images. Looking forward to comments and improvements


akira25 (changes requested):
I personally think, that the folder _packages_ or renamed _imagetypes_ should be named _packagelists_ or something similar. That name would meet better the expectations on the content

SvenRoederer:
Thanks for this comment, but in a16fbfb468e7cdd2bf0e491ff40e6a379542efb2 this folder gets removed anyway

SvenRoederer:
> I personally think, that the folder _packages_ or renamed _imagetypes_ should be named _packagelists_ or something similar. That name would meet better the expectations on the content

@Akira25 I put your suggestion into a separate PR #766

SvenRoederer:
Sadly I must agree to the comment https://github.com/freifunk-berlin/firmware/pull/825#issuecomment-653529148 of @Akira25, but this PR is a working solution around issue #696. And it's the only solution I've seen so far.

Akira25:
I see some great advantages in this commit. Unfortunatly, I wasn't able to test out by building some images. When I just fired some plain `make` command, the build system asked for a `site.conf`. @SvenRoederer is there a readily avaiable site.conf for Berlin?

```
package-src: hostapd-mini  
packages   : hostapd-mini
Makefile:75: *** No site configuration was found. Please check out a site configuration to /home/martin/test_gluon_framework/site.  Stop.
```

SvenRoederer:
I have a (hopefully still working) one at https://github.com/SvenRoederer/ffberlin-gluon-site.
But the my goal is to not rely on a site.conf finally.

Akira25:
Okay, I did checkout this branch and checked out your site-configuration to the directory `./site`. The build looked very promising, but stoped with that prompt:

```
make: *** No rule to make target '/home/martin/test_gluon_framework/configs/common.config', needed by '/home/martin/test_gluon_framework/openwrt/.config'.  Stop.
```

SvenRoederer:
I assume you just called make w/o an explicit goal, right?

I did not change the default goal yet. So have a look at the diff of the Makefile. Calling 'make gluon-compile' and 'make gluon-images' should do the trick.

PR #766, opened 2020-03-07
Rename packages-folder

SvenRoederer:
Rename the folder "packages" to "packagelists". This change will
* avoid the clashes of downloaded feed-repos with our packagelists
* give the folder a more intuitive name (https://github.com/freifunk-berlin/firmware/pull/765#pullrequestreview-370781837)


Akira25:
Ah. That went quick. ^^
Is it appropriate to rename that folder now? I think we should postpone that rename until your suggestion for restructuring the build system is RFC. If done then, we should maybe rebase it into one commit.

@pmelange @booo What do you think?

SvenRoederer:
the commits have been around in this branch for some time, but I did not make a PR out of them before :-)
Probably it's handy, to keep the .gitignore change separate, but squashing is fine.

Why postponing? Currently there ist the problem, that "git status" complains on untracked files in "packages" as soon as "make prepare" was called.

Akira25:
The build of this branch was successful on my machine, but i didn't test the images.

SvenRoederer:
thanks for a separate test, quite sure the images are fine, as the imagebuilder is called with the same arguments.

I'd like to merge this soon.

SvenRoederer:
let's agree on squashing the 1st and the last into one and keep the other two separate, as they don't strictly belong to the folder-renaming.

Akira25:
Okay. I'm fine with that.

Issue #780, opened 2020-04-09
buildenv: using an forked openwrt repo and branch from github fails


pmelange:
I would like to use a forked repo and a test branch to compile a test version of openwrt.  Unfortunately it does not work.

Here are the errors
```
rm -f /home/pmelange/src/mt7621/openwrt/.config                                
rm -f .stamp-
rm -f .stamp-gluon-module-openwrt
echo >>.stamp-gluon-module-openwrt "OPENWRT_REPO=https://github.com/pmelange/openwrt.git"
echo >>.stamp-gluon-module-openwrt "OPENWRT_COMMIT=641703b11b7b4e0e17298bbe5386ae302d9ed955"
git clone $(grep _REPO .stamp-gluon-module-openwrt | cut -d "=" -f 2) /home/pmelange/src/mt7621/tmp/gluon_.stamp-gluon-module-openwrt                          
Cloning into '/home/pmelange/src/mt7621/tmp/gluon_.stamp-gluon-module-openwrt'...
remote: Enumerating objects: 32, done.
remote: Counting objects: 100% (32/32), done.
remote: Compressing objects: 100% (24/24), done.
remote: Total 504820 (delta 11), reused 16 (delta 8), pack-reused 504788
Receiving objects: 100% (504820/504820), 175.36 MiB | 4.92 MiB/s, done.        
Resolving deltas: 100% (344264/344264), done.                                  
cd /home/pmelange/src/mt7621/tmp/gluon_.stamp-gluon-module-openwrt; git name-rev --refs openwrt-* $(grep _COMMIT /home/pmelange/src/mt7621/.stamp-gluon-module-openwrt | cut -d "=" -f 2) | cut -d / -f 2 | cut -d \~ -f 1 >branchname.txt     
cd /home/pmelange/src/mt7621/tmp/gluon_.stamp-gluon-module-openwrt; grep -q undefined branchname.txt  || \
        printf >>/home/pmelange/src/mt7621/.stamp-gluon-module-openwrt "OPENWRT_BRANCH=%s\n" \
                $(echo  | tr '[:lower:]' '[:upper:]') \                        
                $(cat branchname.txt)
rm -rf /home/pmelange/src/mt7621/tmp/gluon_.stamp-gluon-module-openwrt         
rm -f /home/pmelange/src/mt7621/modules
cat >>/home/pmelange/src/mt7621/modules .stamp-gluon-module-openwrt            
[...]

Issue #772, opened 2020-03-27
Build process fails: nchname.txt.txt: No such file or directory


daftmilk:
Hello!

Unfortunately there is no precompiled image for my router:
http://buildbot.berlin.freifunk.net/buildbot/stable/1.0.5/ar71xx-generic/

I am trying to build version 1.05 to update from Freifunk Berlin Hedy 1.0.4 v1.0.4 on Mac Os 10.15.3

The build fails with:

```
cd XXXXX/tmp/gluon_.stamp-gluon-module-openwrt; git name-rev --refs openwrt-* $(grep _COMMIT XXXXX/.stamp-gluon-module-openwrt | cut -d "=" -f 2) | cut -d / -f 2 | cut -d \~ -f 1 nchname.txt.txt
cut: nchname.txt.txt: No such file or directory
make: *** [.stamp-gluon-module-openwrt] Error 1
```

It looks like it would be pretty easy to fix, but I just don't know where to look.

SvenRoederer:
this relates to 95dd252b34e79057a284bc895d85809752cd4147; #738; #739; #740 

> 
> Unfortunately there is no precompiled image for my router:
> http://buildbot.berlin.freifunk.net/buildbot/stable/1.0.5/ar71xx-generic/
> 

Have you tried using the imagebuilder, as per https://github.com/freifunk-berlin/firmware/blob/master/README.md#customizing-make.
Sadly the precompiled imagebuilder has been deleted on download-server, but you can still find it at http://77.245.36.54:8000/firmwares/stable/ar71xx-generic/

SvenRoederer:
> The build fails with:
> 
> ```
> cd XXXXX/tmp/gluon_.stamp-gluon-module-openwrt; git name-rev --refs openwrt-* $(grep _COMMIT XXXXX/.stamp-gluon-module-openwrt | cut -d "=" -f 2) | cut -d / -f 2 | cut -d \~ -f 1 nchname.txt.txt
> cut: nchname.txt.txt: No such file or directory
> make: *** [.stamp-gluon-module-openwrt] Error 1
> ```

On first view it looks like a MacOS-specific Escaping-issue of the master branch. Seems you did not checkout the "v1.0.5" tag.

SvenRoederer:
> On first view it looks like a MacOS-specific Escaping-issue of the master branch.

PR #781 will fix this also.

@daftmilk did you check out the correct branch before you started the build?

SvenRoederer:
[...]

Issue #777, opened 2020-04-07
buildenv: using src-git local_feed_name https://example.com/repo_name/something.git;branch_name broken


pmelange:
One of the supported ways to use feeds.conf includes formatting in the following way:

```
src-git local_feed_name https://example.com/repo_name/something.git;branch_name
```

This, unfortunately, is not working.  The resulting error is:

```
fatal: unable to access 'https://github.com/openwrt/packages.git;master/': The requested URL returned error: 400
```

see https://openwrt.org/docs/guide-developer/feeds#feed_configuration

SvenRoederer:
This is also caused by the "glue-code" to create the modules-file from our feeds.conf, as written in https://github.com/freifunk-berlin/firmware/issues/780#issuecomment-611788382.

PR #781solves issue #780, but not sure if this will also handle the idea of the issue to use the HEAD of a repo.

PR #781, opened 2020-04-10
Use modules-file instead of generating it from config.mk and feeds.conf

SvenRoederer:
The modules-file controls which code of the upstream-repos is used. With the implementation of PR #714 we added some "glue-code" to create the ```modules``` file from the currently used files (```config.mk```, ```feeds.conf```). 
This "glue-code" works well with the official Upstream-repos but seems to have some problems with alternative repos as shown by issue #780. This PR drops the "very brittle" code and moves the repo-definitions to ```modules``` directly.

* Drops the additional converting code
* speeds up build, by not copying each repo an additional time to find the branch of the referencing commit
* moves the definition of any upstream-repo into a single file (not spread around two files anymore)
* an instant lookup of the branch, the code firmware-build is based on, is possible


pmelange:
I am currently building with these changes cherry-picked into a private branch of the "next" branch.  The goal is to prototype kernel 5.4 on the erx-sfp.  

One thing that I noticed, is that I have to put a commit ID into the modules file.  If not, I get the following error:

```
Initialized empty Git repository in /home/pmelange/src/mt7621/openwrt/.git/     
remote: Enumerating objects: 2, done.                                           
remote: Counting objects: 100% (2/2), done.                                     
remote: Compressing objects: 100% (2/2), done.                                  
remote: Total 488150 (delta 0), reused 0 (delta 0), pack-reused 488148          
Receiving objects: 100% (488150/488150), 171.31 MiB | 5.21 MiB/s, done.         
Resolving deltas: 100% (332298/332298), done.                                   
From https://github.com/vDorst/openwrt                                          
 * branch            ramips-5.4 -> FETCH_HEAD                                   
fatal: Not a valid object name: ''.                                             
unable to find commit "" on branch "ramips-5.4" in repo "https://github.com/vDor
st/openwrt.git".                                                                
make: *** [Makefile:90: .stamp-pre-patch] Error 1                               
```

It would be really cool if this code would use the HEAD of the branch if no commit-id is given.

So, I would say that this fixes issue #780, but not #477

pmelange:
Running "make clean" deletes the modules file.  Afterwards running make again results in
```
rm -f /home/pmelange/src/mt7621/openwrt/.config
rm -f .stamp-
make: *** No rule to make target '/home/pmelange/src/mt7621/modules', needed by '.stamp-pre-patch'.  Stop.
```

Also, it seems like that if I change the commit-id (for example in luci) in the modules file, and run make again, then the new commits are not integrated and built.

SvenRoederer:
fixed in 3443833

SvenRoederer:
Any additional feedback?
What about merging this?

pmelange:
I still haven't gotten around to testing alternative git repos. Specifically I would like to try the equivalent of src-lnk, which I have found to be very useful for prototyping

Akira25:
I havn't test this, but I like the goals to be achieved by this PR. Sounds like a good Idea.

pmelange:
I'm having trouble being able to use alternative git feed types.  How are we supposed to be able to use the different types of feeds support by openwrt and shown here: https://openwrt.org/docs/guide-developer/feeds

SvenRoederer:
this PR is only about solving the problem described in #780 (interpreting external GIT repos from config.mk and feeds.conf to build the modules-file). In addition this PR should also fix #772. 
We should avios to use a solution for an issue (here #780) to solve another different issue (#777 ) too. Even both issues somehow relate to each other.

SvenRoederer:
> I'm having trouble being able to use alternative git feed types. How are we supposed to be able to use the different types of feeds support by openwrt and shown here: https://openwrt.org/docs/guide-developer/feeds

@pmelange can you put this into a separate issue?

pmelange:
> > I'm having trouble being able to use alternative git feed types. How are we supposed to be able to use the different types of feeds support by openwrt and shown here: https://openwrt.org/docs/guide-developer/feeds
> 
> @pmelange can you put this into a separate issue?

See issue #789

SvenRoederer:
Independent of fix for issue #789 this PR will solve the #780 & #772 already.

SvenRoederer:
I'm going to merge this by this week, as it solves to issues and doesn't break anything. I tested it for some time in the "SAm0815-experimental" branches.

SvenRoederer:
merged into master with 487cb331c1d48e5f4b092470e7b24032f4691c0b and c6aa875f091a3c08065a2e960729d422a6fda122.

PR #782, opened 2020-04-10
ReadMe: add contact-info to the github-team

SvenRoederer:
We created a Github-team (@freifunk-berlin/firmware) for the firmware-group, so we add this to the contact-section.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,8 @@ More user relevant information about the firmware are on the wiki page at: https
 * [ReleaseNotes](https://wiki.freifunk.net/Berlin:Firmware/v1.0.2)
 * a tutorial ([en](https://wiki.freifunk.net/Berlin:Firmware:En:Howto) / [de](https://wiki.freifunk.net/Berlin:Firmware:Howto)) on router configuration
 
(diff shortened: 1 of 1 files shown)


db4rne:
Its not a big change, but it might be useful to add the berlin freifunk contact page as well. https://berlin.freifunk.net/contact/

SvenRoederer:
As the "berlin freifunk contact page" is more end-user related, I'm not sure if this is really a good resource for developer, which will be on Github.

pmelange:
Looks fine to me too.

But I also agree with @db4rne that having a link to https://berlin.freifunk.net/contact/ might still be useful.  Maybe make this section "Contact/More Info" section into two parts, one for users and one for developers, would be good anyhow.  But another change like this could also go into another PR in the future.

pmelange:
The tutorial is also there already, but that is for end-users and not just for developers.  The links to the tutorials would, in this case, belong to the user part of the subsection I described in the last comment.

Akira25:
I'll merge this, as there seems to be a consensus on this. For the further suggestions on the contact form, I'll create another MR.

PR #785, opened 2020-04-16
Fix current builderrors (package-conflict, bad umask)

SvenRoederer:
As the buildbot is online again, we see 2 issues 
* imagebuilder fails as of bad umask
  * this happens in worker wg1337
  * error: 
> Build dependency: Please build with umask 022 - other values produce broken packages
* build fails in openwrt-recipe "pacakge/install"
  * verbose build shows: 
> * check_data_file_clashes: Package kmod-ath10k-ct-smallbuffers wants to install file /var/lib/buildbot/workers/ff-b/build/ar71xx-mikrotik/build/firmware/openwrt/build
        But that file is already provided by package  * kmod-ath10k-ct
>  * opkg_install_cmd: Cannot install package kmod-ath10k-ct-smallbuffers."

---
* umask-issue can be fixed by setting it proper at runtime
* package-dependency is caused by have boards with default- and smallbuffers-variant sharing the same filesystem
---

I suggest to rebase  these commits on master, rather than doing a merge commit. The rebase looks more nice.


booo:
Looks good. Should we merge?

Should we ask the buildbot worker maintainer to fix the umask?

SvenRoederer:
I had seen the "package-dependency"-issue some time on my local builds already, but wasn't sure if this is a local problem only. The buildbots prove me, that it's not local.

umask
* we can reconfigure the buildbot, but this will not fix the issue when building local (see issue https://github.com/freifunk-berlin/firmware/issues/431)
* to change on the buildhost, it seems more elegant to do this via the buildmaster-script. This is a central place and not every (new) buildhost needs to be touched

pmelange:
Looks good to me too.  If for some reason we have problems in the future, we can always make other changes.  Rebased-merging

pmelange:
The first build with these changes failed.  It might be because of an unrelated problem.

https://buildbot.berlin.freifunk.net/#/builders/2/builds/27

SvenRoederer:
> The first build with these changes failed. It might be because of an unrelated problem.
> 
> https://buildbot.berlin.freifunk.net/#/builders/2/builds/27

This is very likely caused by Low disc-space (https://github.com/freifunk-berlin/firmware/issues/602)

Issue #747, opened 2019-12-18
change versioning scheme to based on release date


SvenRoederer:
Some points should make us think of changing the current versioning-scheme.
* upstream removal of "CONFIG-Option VERSION_NICK" (#515) which makes it hard to find a proper place to store our Releasename (Kathleen, Hedy, Gonzo). The way choosen for "Gonzo-rc"-builds was to combine the numeric version with the releasename, which instantly caused such issues --> #710 
* Semantic versioning: on the [Mailinglist there was some discussion](https://lists.berlin.freifunk.net/pipermail/berlin/2019-June/039493.html) if it is really handy for enduser-software. Modifying this scheme, from what was agreed in #145, and dropping the releasename will avoid confusion as we have seen when going from Kathleen to Hedy. Even I support the idea of honoring famous (female) persons of the IT-sector.

I vote for going to something like "Freifunk Berlin 2020.02.0" for a release scheduled for early 2020. Consequently the 1st maintenance-release will become "Freifunk Berlin 2020.02.1" and a prerelease will be "Freifunk Berlin 2020.02.0-rc1". This will not strictly meet the SemVer-definition, as we will change the major-number on a annual basis, even if there are no incompatible API changes.

everloop2:
numbers: appreciate it. Its far more easy to compare.

znrR:
Pro Nickname: "Marketing" because we can dedicate a major release to an "influentual" person such as Hedy Lamar
Con Nickname: The field to store that nickname has been removed by OWRT and it would mean we need to attach it to some other field (eg. vendor) which requires follow up work on various currently unknown non firmware related projects to hide that information again.

znrR:
proposal: could also be placed in the footer and no longer be part of the "real" version number 
![Screenshot 2020-04-21 at 20 45 18](https://user-images.githubusercontent.com/59607366/79902005-07e54d80-8411-11ea-88f0-5c3a54329f38.png)

znrR:
technically this would look like:
$CONFIG_VERSION_DIST Firmware Release $CONFIG_VERSION_NICK! $CONFIG_VERSION_DEDICATION

znrR:
agreed as to do for @SvenRoederer

znrR:
also we agreed on changing the versioning from a consecutive 1.1.XX to timestamp based 2020.XX.Y which is similar to the OWRT scheme.

PR #792, opened 2020-04-29
embedded_files: add freifunk favicon to embedded_files

### LINKED ISSUE #791: Enhancement: Use the freifunk favicon.png for the luci interface
I recommend using the freifunk favicon for the web interface. 

favicon

To do this, I would recommend putting this into the $OPENWRT_DIR/files/www/luci-static/bootstrap folder to overwrite the original.

Screenshot_2020-04-29_12-17-33

Original size: 733 bytes
Freifunk site:1528 bytes (possibly this can be optimized).

pmelange:
This overwrites the existing openwrt favicon with the freifunk favicon.

Compiled and tested on a WDR3600.  The favicon come up as expected.

Fixes #791 

Signed-off-by: pmelange <isprotejesvalkata@gmail.com>


booo:
Can you please check if we can serve a SVG instead of a PNG?

pmelange:
It would also work with a SVG, but in order to do that we would need to modify/patch https://github.com/openwrt/luci/blob/81effc111227bc65cb974829f17a45515fbe9807/themes/luci-theme-bootstrap/luasrc/view/themes/bootstrap/header.htm#L29

pmelange:
As an extra bonus, I tried renaming the SVG to favicon.png to see if it works.  Nope.  Instead no favicon is displayed.

SvenRoederer:
As we spoke on modifying the page-footer, to include a dedication-note (#747) I suggest to use a LuCI-theme for this. Both changes can be added on a clean and standard way, which IMHO is much better than hacking this via "embedded_files".

Similar had been done in "officially" in LuCI (before some packages have been moved to the freifunk-repo (https://github.com/freifunk/openwrt-packages/tree/ac16b4a424b5761950a7b8d7fd7b439fe334e58d/themes/luci-theme-freifunk-generic).

pmelange:
> better than hacking this via "embedded_files".

I do not agree that this is a hack, as it is a documented and support way to customize an installation: https://downloads.openwrt.org/releases/18.06.4/targets/ar71xx/generic/openwrt-18.06.4-ar71xx-generic-ubnt-unifiac-mesh-squashfs-sysupgrade.bin

If/when a desired theme has been created, then this commit could be reverted.

Otherwise, please just NACK and close this PR.

pmelange:
Oops, copied the wrong link into the text.  I have updated the comment.

https://openwrt.org/docs/guide-developer/build-system/use-buildsystem#custom_files

SvenRoederer:
Please check the new Freifunk-generic-theme (https://github.com/freifunk/openwrt-packages/pull/30)

SvenRoederer:
the favicon of this branch has been integrated into freifunk/openwrt-packages#30

SvenRoederer:
Another drawback of using embedded_files for this: it will only work for the specific theme "bootstrap" and not for others.

PR #793, opened 2020-05-07
README.MD: Build Prerequisites: Arch/Antergos: Build dependency: python

everloop2:
Build dependency: Please install Python 2.x

`sudo pacman -S python` -> installs python3.x

`sudo pacman -S python2` -> installs python2.x

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,7 +45,7 @@ zypper install git ncurses-devel zlib-devel gawk time \
 On Arch/Antergos:
 ```
 pacman -S base-devel git ncurses lib32-zlib gawk time unzip perl-xml-libxml \
- flex wget gettext quilt python openssl
+ flex wget gettext quilt python2 openssl
 ```
 
 ### Building all firmwares


booo:
Could you provide some information why python 2 is a requirement for the build process. Python 2 is end of life and we don't want to support end of life software.

SvenRoederer:
https://github.com/openwrt/openwrt/blob/81f3f6540e66e21be877b99e6524ff91bcea1805/include/prereq-build.mk#L132 : still a build-requirement for OpenWrt-19.07

everloop2:
followed https://github.com/freifunk-berlin/firmware#build-prerequisites

installed "needed" build packages

started building on arch -> it checks for dependencies and says:

> Build dependency: Please install Python 2.x

did `sudo pacman -S python2`

now its building https://github.com/freifunk-berlin/firmware master ;)

dunno if python2 is essential

SvenRoederer:
Seems valid for our master-branch (based on OpenWrt-19.07), so merged in 52e287d377212f491b2e51313946c793f01719d2. Also backported to Hedy-1.0.x.

OpenWrt-master (our next-branch) don't need python2 anymore, but I expect that we need to update the SuSE and debian build-dependencies in the README for these distros.

SvenRoederer:
> ... is end of life and we don't want to support end of life software.

@booo then lets quickly push a new release based on OpenWrt-19.07 or 18.06, as our current releases are still based on lede-17.01, which is [EOL since June 2019](https://openwrt.org/releases/17.01/notes-17.01.7).

PR #794, opened 2020-05-09
build: add make-goal "update-patches"

SvenRoederer:
This PR ports the "update-patches" reciepe from Gluon (https://gluon.readthedocs.io/en/v2020.1.x/dev/basics.html#working-with-repositories).
With calling "make update-patches" the changes to the upstream repos will be compared against the base-commit in ```modules``` and every commit will be converted into a patch. This is especially useful when adding a new patch or updating an existing patch/commit which doesn't apply any more (as of changed base-commit).
As this recipe will not be called automatically, there is no change to the current normal build-workflow.

tested in: http://buildbot.berlin.freifunk.net/#/builders/1/builds/80


SvenRoederer:
based on closing of #790: we should also drop the patches/patches.rst file.

SvenRoederer:
Tested this intensively on my usual work without issues and will merge soon.

SvenRoederer:
* manually merged in 4ea7376c49
* "make update-patches" called in ac729e751669

PR #796, opened 2020-05-11
RFC: simplify Makefile by using a separate Makefile for buildbot

SvenRoederer:
This change is an idea to simplifiy the Makefile. 
In our current Makefile are several test for running on a buildbot to tweak some settings. This includes:
* adding the special ```configs/common-autobuild.config```
* trying to free some diskspace during the build

The idea is:
* to have the current Makefile as it is for individual builds
* to have a separate Makefile (```Makefile.buildbot```), which explicitly calls the recipes of the normal Makefile
  * this Makefile is directly called by the buildbots using ```make -f```
  * this way we can remove the openwrt folder completely after generating the .ipks, imagebuiler and sdk (as shown by the provided example)
  * allows easy integration of buildbot specific tweaks, e.g. parallel execution of imagebuilder to speed up the build, upload of buildlogs, ...

I added this to the "SAm0815-experimental" builds, to test it.

---

The shown PR uses an intermediate Makefile which tweaks the process to use the Makefile.buildbot when running on the buildbot (based in the IS_BUILDBOT flag). The final version will have the ```Makefile.buildlocal``` still as original ```Makefile```.


SvenRoederer:
This might help to fix the "out if disc space" errors seen of the buildslaves (https://github.com/freifunk-berlin/firmware/issues/602)

SvenRoederer:
integrated into PR #839, so closing this one.

Issue #801, opened 2020-05-18
wizard: "tab" symbols on to of page


SvenRoederer:
### Bug report

**What is the problem?**
Since OpenWrt 19.07.3 was released there are some "tab"-symbols below the page-header. These "tabs" don't have any labels, but can be used to directly jump to any configuration-step directly.
![wizard_openwrt-19 07 3](https://user-images.githubusercontent.com/1755362/82242651-81079000-993e-11ea-9164-ebc493cf09a3.png)

**What is the expected behaviour?**
This was not seen in OpenWrt-19.07.2 based builds (up to 98eab01cbf3cc9cb74f00b0e17942197432381cf)

![wizard_openwrt-19 07 2](https://user-images.githubusercontent.com/1755362/82242674-8bc22500-993e-11ea-8f04-cda5f35859b9.png)

**Firmware Version:** a333cd508e65a652af6f351b0c3e7a1249061535 ; just after upgrading (sysupgrade -n) on the initial page.

SvenRoederer:
Our patch "luci/split-luci-base_PR2817.patch" is not affecting this issue.

PR #802, opened 2020-05-18
update to Openwrt 19.07.4

SvenRoederer:
update master to just released <s>openwrt-19.07.3</s> openwrt-19.07.4, and update to HEAD of our packages feed.
We should merge even it will cause Issue #801, but what options do we have than fixing our wizard-code.

runtime-tested on ramips-mt76x8, ramips-mt7620 (ath79-generic)

EDIT: updated to OpenWrt-19.07.4, which has been released in the mean time.


everloop2:
System: Linux 5.7.14-1-MANJARO (x86_64)

tried to compile and run into:

1. seems build doesnt like multithreading - various build breaks: `make -j1` does help

2. Failed to build miniperl. Please run make minitest.

seems to be fixed in https://github.com/openwrt/packages/pull/12391

> The perl Configure file was matching GCC 10 against "1*" and treating it
> as GCC 1, causing ABI breakage and segfaults.
> 
> Cherry-pick the upstream patch which fixes it to check against (e.g)
> "1.*" instead, which will make it work for hundreds more GCC versions
> to come.

Perl/perl5@6bd6308

SvenRoederer:
@everloop2 seems you are always using to beeding edge OS.
Luckily the patch you mentioned was backported from master (https://github.com/openwrt/packages/commit/65578a43f0d12c02888df00b6fdc90c73a02875c) to openwrt-1907 (https://github.com/openwrt/packages/commit/445c3d8f86b85c2c6fc4a8d7a003b06ca219ffb1). So as it fails to compile because of the seen failure, even the targeted release 19.07.3 is too old to include this fix.
I assume using https://github.com/SvenRoederer/freifunk-berlin-firmware/commit/7bf49639c2ce6ace48337f9c15df7f5e2439e404 (monthly rollup up to end of July 2020) will build. 

Anyway, I in favor of forgetting of OpenWrt-19.07 and to jump to upcoming OpenWrt-20.x directly (https://github.com/freifunk-berlin/firmware/issues/834).

SvenRoederer:
Just rebased on master and updated to fresh OpenWrt-19.07.4.

I assume it has still the Web-UI issue #801, mentioned in the PR-message, but at least brings additional security-updates. It will probably also fix the build-issue mentioned by @everloop2 in https://github.com/freifunk-berlin/firmware/pull/802#issuecomment-675483313

everloop2 (approved):
System: Linux 5.8.6-1-MANJARO (x86_64)

-> builds flawless - multithreading enabled:

`make -j4 V=s TARGET=lantiq-xrx200 PACKAGES_LIST_DEFAULT=notunnel_xDSL`

-> runtime tested: FritzBox7360 (mesh only)

SvenRoederer:
Thanks for checking.
Do we have a lantiq-xrx200 target already?

Regading luci-proto-ppp and irqbalance: 
* since some time I've this patch included https://github.com/SvenRoederer/freifunk-berlin-firmware/commit/b78f362b3acaf4fcc7ddf1339e553f3352b2e728 to build irqbalance.
* luci-proto-ppp is required for the DSL-port, which is WAN?
* both packages seem hardware-related (per board), which can not be well mapped in our build-system. To also enable board-specific packages I had adapted the Gluon-build-system. This offers a way to implement this. See https://github.com/freifunk-berlin/firmware/pull/765 (see "packages =" of https://github.com/freifunk-berlin/firmware/blob/gluon-scripts/targets/ath79-generic)

everloop2:
> Thanks for checking.
> Do we have a lantiq-xrx200 target already?

yes: Fritzbox 7360(sl)(V2) and have a FritzBox3370 runnig
> Regading luci-proto-ppp and irqbalance:
> 
> * since some time I've this patch included [SvenRoederer@b78f362](https://github.com/SvenRoederer/freifunk-berlin-firmware/commit/b78f362b3acaf4fcc7ddf1339e553f3352b2e728) to build irqbalance.

-> fix irqbalance directory not created -> https://github.com/openwrt/packages/issues/12377 -> "fixed" by: https://github.com/openwrt/packages/commit/4f0c847828b9602d12d13059a5e8bb159120dcb7, somehow "irqbalance" directory is not created :/
workaround: firmware/openwrt/package/base-files/files/etc/init.d/boot -> add "mkdir -p /var/run/irqbalance"
-> add irqbalance autostart -> /firmware/openwrt/package/base-files/files/etc/rc.local -> /usr/sbin/irqbalance
-> to check if interrupts are distributed to 2nd core: `cat /proc/interrupts` -> never seen irqbalance distribute interrupts other then CPUcore ones o.O
> * luci-proto-ppp is required for the DSL-port, which is WAN?

yes: without WAN is not editable - if firmware is a default "freifunk" build missing packages can be installed via opkg (for my builds i include them manually for now)
> * both packages seem hardware-related (per board), which can not be well mapped in our build-system. To also enable board-specific packages I had adapted the Gluon-build-system. This offers a way to implement this. See #765 (see "packages =" of https://github.com/freifunk-berlin/firmware/blob/gluon-scripts/targets/ath79-generic)

This will do the trick ;)

SvenRoederer:
regarding luci-proto-wan and irqbalance: I'm not sure if we should add this into the generic packagelists, as both are more board-specific.
Can you open a new issue about handling such packages? So we can find a solution for this and don't have to kidnap this one here.

SvenRoederer:
> System: Linux 5.8.6-1-MANJARO (x86_64)
> 
> -> builds flawless - multithreading enabled:

@everloop2  Do you have permissions to merge? (I assume you have) So will you do the merge if it's looks good for you?

everloop2:
> > System: Linux 5.8.6-1-MANJARO (x86_64)
> > -> builds flawless - multithreading enabled:
> 
> @everloop2 Do you have permissions to merge? (I assume you have) So will you do the merge if it's looks good for you?

Yes it shows Squash and merge - did not do it cause its not my pull request and so: "nobody can blame me" xP

> regarding luci-proto-wan and irqbalance: I'm not sure if we should add this into the generic packagelists, as both are more board-specific.
> Can you open a new issue about handling such packages? So we can find a solution for this and don't have to kidnap this one here.

Yep, will open issue for WAN_DSL boxes like Fritzboxes.

SvenRoederer:
> Yep, will open issue for WAN_DSL boxes like Fritzboxes.

--> #838

PR #805, opened 2020-05-28
GHA/pr-conflict: add workflow to automatically find conflicting PRs

SvenRoederer:
Adds an GithubAction that checks if any open PR can not be merged on master
branch after a new commit was made to the master branch. If such conflict
will be detected, the affected PR will get the label "has conflicts". If
during a following run a conflicting PR becomes mergable, the flag will be
removed again.
This action is based on https://github.com/marketplace/actions/auto-label-merge-conflicts

This requires the correct label "has conflicts" needs to be created or the actions config must be adapted.

PR #807, opened 2020-06-07
Misc fixes without functional change for enduser

SvenRoederer:
This fixes should be just rebased onto.

The removed boards are evaluation-boards, so they will not be used by end-users anyway.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,7 +2,8 @@
 .stamp*
 patches/*~
 bin
-^openwrt
+/openwrt
 firmwares
 dl
 /packages
+/tmp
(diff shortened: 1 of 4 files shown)


SvenRoederer:
Rebased after seeing this PR went into conflict.  Having PR #805 merged would have been a great help to get an automated notification.

SvenRoederer:
@pmelange Why have you squashed these 4 commits, each dealing with a very different issue in the code, in to single one? This way it becomes more difficult to read the history. Now the commit-headline doesn't say anything meaningful and the interested developer has to read the full commit message. Even for me as creator of this PR it's hard to read it and "desquash" it

pmelange:
That was not my intention.  Sorry.  Feel free to revert the squashed commit and merge the others.

SvenRoederer:
pushed one revert and the 4 individual commits (19c8a1393c5619b2e0b499df10c9684ae7630e39)

Issue #722, opened 2019-08-14
 TP-Link CPE210 Ver: 2.0/3.0 wifi not working


Mitzlaff:
### Bug report

**What is the problem?**
WLAN does not work. 

When using LAN connection I see this:

![grafik](https://user-images.githubusercontent.com/2312815/63012283-43934500-be8a-11e9-9a27-0eabac92de51.png)

**What is the expected behaviour?**
SSID showing up when scanning for configured SSID name (freifunk-backyard)

**Firmware Version:**
freifunk-berlin-dev-daily-6431131-tplink_cpe210-v3-sysupgrade.bin

**Site Configuration:**
default, just using Luci

**Custom patches:**
none

SvenRoederer:
Does disabling the Mesh-link (intern-ch13.freifunk.net) fixes it?

everloop2:
see: https://bugs.openwrt.org/index.php?do=details&task_id=2333

https://github.com/openwrt/openwrt/pull/1951#issuecomment-512795229

2-1 week ago searched for wifi bug - other devices using similar hardware seems to affected too

psyborg55:
> -> wifi bug hint: seen somebody mentioned in CPE210 bug threads different irqs

doesn't appear to be a problem, at least according to bootlog https://github.com/openwrt/openwrt/pull/2468#issuecomment-537549406

AR9344 and AR9342 work fine with this irq setup

everloop2:
seems to be fixed:

mac80211 ath9k: force QCA953x clock to 25MHz

https://github.com/openwrt/openwrt/commit/4c6fe32468bc60cc25a8c298498c0be3c73e7378

> The QCA953x only supports 25 MHz refclk, however some OEMs set an
> invalid bootstrap value for the REF_CLK option, which would break the
> clock detection in ath9k.
> 
> Force the QCA953x refclk to 25MHz in ath9k, as this is (according to the
> datasheet) the only valid frequency.

> Signed-off-by: David Bauer <mail@david-bauer.net>

did not test it yet.

SvenRoederer:
> **Firmware Version:**
> freifunk-berlin-dev-daily-6431131-tplink_cpe210-v3-sysupgrade.bin
> 
> ****

For the records: this was the first build based on openwrt-master using the sth79-target for the CPE210-devices.

Mitzlaff:
[...]

PR #827, opened 2020-06-28
add support for TPlink CPE v2,v3 for ath79

SvenRoederer:
These boards are supported on ath79 since April 2019.  So I don't expect any problems hardware-wise.


SvenRoederer:
as per https://github.com/freifunk-berlin/firmware/issues/722#issuecomment-662515298, the Wifi-issues of the CPEv2 (CPEv3) have been fixed upstream and the units working fine.

PR #848, opened 2021-05-02
patches/luci: add "german translation for dismiss"

SvenRoederer:
add LuCI PR https://github.com/openwrt/luci/pull/5024 till it's upstreamed


SvenRoederer:
Upstream change will fix this with a different approach, but we can still use this patch till it's fixed upstream.

SvenRoederer:
has been merged upstream (https://github.com/openwrt/luci/commit/1a5958416747a0cbb137e35c056c6f4161783c28 for 21.02) and can be closed.

PR #849, opened 2021-05-02
Fix faulty LAN-port in TPLink EAP225 Outdoor

SvenRoederer:
Add this patch till it's accepted upstream; see http://lists.openwrt.org/pipermail/openwrt-devel/2021-April/034909.html

the device was recently added to OpenWrt master and OpenWrt-21.02, therefore it depends on the update to OpenWrt-21.02.0-rc1


SvenRoederer:
superseded by 86f2df7 / 25a66a55d61ef72

PR #852, opened 2021-05-07
configs: revert to upstreams idea of SQUASHFS_BLOCK_SIZE (beside tiny devices)

SvenRoederer:
back in e6708dead14 (merge 8b1eb621a652 - "4MB devices") the SQUASHFS_BLOCK_SIZE was set
to 512kb to provide better compression, esp. for 4MB flash devices.
OpenWrt setup the tiny subtarget for such devices, where we keep this special handling as
intended. For all other subtargets it should be safe to revert back to the upstream-default
of this setting (256 kb in April 2021).
This setting is used very inconsistently on our targets at all.


SvenRoederer:
merged in 1b01478efa7b89b5e1f6a71c5ee636f7b63d39ac

PR #854, opened 2021-05-10
modules: download OpenWrt repo from GitHub

SvenRoederer:
use GitHub mirror of official OpenWrt-repo, as it seems more effective
network wise. So we don't have to push all the data trought the Internet,
but just use GitHub internal traffic.


SvenRoederer:
merged by e3fbce166e55a09792f87e0f306adef6ca8ad9a7

Issue #838, opened 2020-09-26
xDSL: WAN not configurable via LuCI due missing packages


everloop2:
**Firmware Version:**

master: Freifunk Berlin Development 91b6ae0 -> think all others too

**What is the problem?**

boxes got a DSL port (FritzBox 7360x, 3370..) -> WAN not configurable via LuCI due missing packages:

```
luci-proto-ppp
luci-proto-ipv6
ppp
ppp-mod-pppoe

```
**What is the expected behaviour?**

WAN should be configurable via LuCI - expecially to switch: DSL_WAN to LAN_WAN

(OpenWRT stable/snapshot needed packages are not missing)

**Workaround/Solutions:**

direct use of DSL_uplink is not done in most cases and [setting it up](https://openwrt.org/docs/guide-user/network/wan/isp-configurations#general_xdsl_configuration_step_by_step) not k.i.s.s. .

1. set WAN default to DHCP -> prevents WAN is not configurable via LuCI due missing ppp_xx packages:

```
/firmware/openwrt/target/linux/lantiq/base-files/etc/board.d/02_network

ucidef_set_interface_wan "$interface_wan" "dhcp"
```
or set it via Freifunk wizard.

2. remove all xDSL related packages - if DSL_uplink is needed, manually download packages and set it up after wizard

/firmware/packagelists/..

```
# add/enable to prevent Luci DSL_WAN no proto bug
-luci-proto-ppp
-luci-proto-ipv6
-ppp
-ppp-mod-pppoe

# remove lantiq xDSL: xway_ADSL (EasyBox803..) and xrx200_xDSL (FritzBox 7360x, 3370..)
-ppp-mod-pppoa
-kmod-ppp
-linux-atm
-ltq-adsl-app
-kmod-ltq-atm-ar9
-kmod-ltq-adsl-ar9
-kmod-ltq-adsl-ar9-mei
-kmod-ltq-adsl-ar9-fw-a
-kmod-ltq-adsl-ar9-fw-b
-kmod-ltq-atm-danube
-kmod-ltq-adsl-danube
-kmod-ltq-adsl-danube-fw-a
-kmod-ltq-adsl-danube-fw-b
-kmod-ltq-adsl-danube-mei
-br2684ctl
-bspatch
-dsl-vrx200-firmware-xdsl-a
-dsl-vrx200-firmware-xdsl-b-patch
-kmod-atm
-kmod-ltq-atm-vr9
-kmod-ltq-deu-vr9
[...]

PR #863, opened 2021-06-27
add Feature "devicepackages"

SvenRoederer:
This extends the `assemble_firmware.sh` script in a way to parse a additional file `profile-packages.txt` which allows to specify packages that should be installed / removed on a per profile / board basis.
The code is based on the changes of @WeimarNetz with minor changes for our flavor of the `assemble_firmware.sh`. Initially it was intended to fix issue #696 around the ath10k-wave1 chips. 
Quickly it turns out that this might also fix issue #838 (additional drivers for boards with integrated xDSL-hardware) and also allows adding mass storage support (SATA, SD-Card) in default images.

### DIFF
--- a/configs/ath79-generic.config
+++ b/configs/ath79-generic.config
@@ -1,6 +1,7 @@
 CONFIG_TARGET_ath79=y
 CONFIG_TARGET_ath79_generic=y
 CONFIG_PACKAGE_ath10k-firmware-qca988x=m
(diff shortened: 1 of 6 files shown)


SvenRoederer:
@weimarnetz @andibraeu This PR is based on you recent changes. Probably some of the change also apply to you, e.g b80e691, da82f73. I also added some board-definitions that are not in your list.

andibraeu:
nice, thank you!

PR #866, opened 2021-07-18
Update OpenWrt 21.02 to state of 1. July 2021

SvenRoederer:
important fixes on the way to OpenWrt-21.02 release

* fixes build for TPLink CPE (rootfs too small)
* fixes network in failsafe mode
* don't update routing-feed for now as of https://github.com/openwrt/routing/issues/691


SvenRoederer:
merged with acf592b1d5dcd2d1cf4f

PR #868, opened 2021-07-25
make ath10k a smallbuffer-variant 

SvenRoederer:
For some ath10-wave1 boads we use the non-ct driver (#696). Test have shown that some boards are not running stable with this. I can confirm for a ubnt_nanostation_ac. This seems to be caused by the "low RAM" these boards have (NanoStation AC has 64MB). That is the same reason OpenWrt made an official "ath10-ct-smallbuffers" package.

This imports the patch that was used on OpenWrt, before they switched to the "ct-"driver, to address the RAM-issue.

As the "Wave1" issue only affects some boards it seems not worth to create a "ath10-smallbuffers" package. So let's use the smallbuffers for all boards, even they have 128 MB (as EAP225-outdoor which is not affected by sudden reboots).

* see https://github.com/FreifunkFranken/firmware/commit/aafd9edd1d2e23c08d2b2325f69e42f000c3fdfe
* cherry-picked from https://github.com/weimarnetz/firmware/commit/e7577717b57d7c608c53e2c5d1b3f8b7f8a34169

### DIFF
--- a/patches/openwrt/0006-mac80211-add-back-ath10k_pci-memory-hacks.patch
+++ b/patches/openwrt/0006-mac80211-add-back-ath10k_pci-memory-hacks.patch
@@ -1,0 +1,73 @@
+From: Adrian Schmutzler <freifunk@adrianschmutzler.de>
+Date: Wed, 10 Feb 2021 22:53:00 +0100
+Subject: mac80211: add back ath10k_pci memory hacks
+
+These hacks have been removed in commit 1e27befe63ff ("mac80211: remove
+ath10k_pci memory hacks").
+
+However, since we still use mainline ath10k, we will need them.
+
+Signed-off-by: Adrian Schmutzler <freifunk@adrianschmutzler.de>
+
+diff --git a/package/kernel/mac80211/patches/ath/960-ath10k-limit-htt-rx-ring-size.patch b/package/kernel/mac80211/patches/ath/960-ath10k-limit-htt-rx-ring-size.patch
+new file mode 100644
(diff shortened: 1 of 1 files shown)


SvenRoederer:
had it running for some days on a NanoStation-AC (64MB RAM) and a TPlink EAP225-Outdoor (128MB RAM). Both devices are configured as STA to the same AP.


## 64MB devices (NanoStation AC)
Monitoring shows for the unpatched NS-AC average uptime of only some hours, with a lot of traffic loss as of crashed olsrd.
NSM-ac (regular ath10k)
Same setup with the patched "smallbuffer" driver on the NS-AC - runnig for 4days without reboot or crash of olsrd. NSM-ac (smallbuffer ath10k)

## 128MB devices (TPLink EAP225 Outdoor v1)
There are no indication in the monitoring, that any performance-parameter has changed.

## Summary

As we use the non-ct driver only for 802.11s / STA, which usually have a low client-count, I don't expect any performance problems with smaller buffers on 128MB or 64MB devices. For AP-use it might be different, but was not tested. 
But using this patch allows reliable use of 64MB devices, whereas will for sure not make AP-situations reboot randomly. So let's follow @FreifunkFranken and @weimarnetz

SvenRoederer:
@adschm probably you remeber best, why OpenWrt did not make a separate ath10-smallbuffers package as it has been done for ath10-ct.

SvenRoederer:
merged by 7c33f06

adschm:
> @adschm probably you remeber best, why OpenWrt did not make a separate ath10-smallbuffers package as it has been done for ath10-ct.

Nobody volunteered. And it's not entirely trivial. If you are interested, feel encouraged.

PR #870, opened 2021-08-01
make 32MB RAM devices usable again

SvenRoederer:
As stated by OpenWrt there is limited support for devices with less than 32MB RAM. With out firmware we can see boards with 32MB run very unstable, as of frequent OoM issues.
This changes removing some features, from packages and kernel, to free up some RAM in order to improve the stability. This will give the users some time to upgrade their hardware but keep recent and more secure code than LEDE based "Hedy".

Runtime-tested on NSM2-loco


SvenRoederer:
There was some discussion on the mailing list (https://lists.berlin.freifunk.net/pipermail/berlin/2021-September/052634.html). Beside the usual "let's upgrade RAM and Flash-chips", there are also some replies being interested in reusing the current hardware.
With this change it's also easy to support rechiped (16/64MB) boards, which can directly be added into the ath79-generic target, with a changed board-id.

merged with ec617b18f3

